PR #33, opened 2026-05-16
fix(web): use app url for MCP endpoint

### LINKED ISSUE #27: fix(ui): use NEXT_PUBLIC_APP_URL for the MCP endpoint shown in Settings → API Keys
## What's wrong

The **Settings → API Keys** page shows the MCP endpoint URL that users
copy into their Claude / Cursor / Zed config. It builds the URL from
`window.location.origin`, which causes two problems:

1. **SSR + hydration drift.** The page is server-rendered first with
   no `window` available, then re-rendered on the client. With browser
   caching, service workers, or a slow hydration, the URL the user
   ends up seeing can be wrong — including a stale `http://localhost:3000`
   if they ever opened the same page during local development.
2. **Inconsistent with the rest of the codebase.** Other places that
   need the canonical app URL read it from
   `process.env.NEXT_PUBLIC_APP_URL` (see references below). This one
   spot drifted to `window.location.origin` and now shows a different
   value than every other URL-aware feature in the app.

End result: the user can copy the wrong URL into their MCP client
config and nothing connects.

## Expected

The MCP endpoint URL should be built from
`process.env.NEXT_PUBLIC_APP_URL`, falling back to
`window.location.origin` **only** when the env var isn't set (graceful
fallback for setups where it's been left blank).

## Where to fix

`web/src/components/settings/api-keys-section.tsx` — two spots use
`window.location.origin`:

- Around line 144: the `<code>` element that displays the URL
- Around line 152: the `copy()` button's `onClick` (same string)

Both should switch to the env-var-first pattern.

## Reference pattern

The rest of the codebase already does it this way:

- `web/src/app/api/stripe/checkout/route.ts` (~line 69)
- `web/src/app/api/stripe/success/route.ts` (~line 8)
- `web/src/lib/actions/team.ts` (~lines 208, 273)
- `web/src/app/layout.tsx` (~line 30)

All do roughly:

```ts
const appUrl = process.env.NEXT_PUBLIC_APP_URL || /* fallback */;
```

A small helper or a `useMemo` inside the component is fine. Keep it
small.

## How to verify

Open `/dashboard/settings` → **API Keys** tab. The MCP endpoint URL
shown should match whatever `NEXT_PUBLIC_APP_URL` resolves to in your
environment. Click the copy button — clipboard should contain the same
URL.

## Contribution guide

See [CONTRIBUTING.md](../CONTRIBUTING.md) for branch naming, commit
conventions, and PR workflow.

akmhatey-ai:
## Summary

- Use `NEXT_PUBLIC_APP_URL` as the canonical base for the Settings API Keys MCP endpoint.
- Reuse one computed endpoint string for both the displayed URL and copy button.

## Context

Fixes #27.

The settings page was using `window.location.origin` directly for the MCP endpoint, while the rest of the app uses `NEXT_PUBLIC_APP_URL` for canonical app URLs.

## Changes

- Added a small MCP endpoint helper in `api-keys-section.tsx`.
- Normalized a trailing slash from `NEXT_PUBLIC_APP_URL`.
- Kept `window.location.origin` as the fallback when the env var is not configured.

## Testing

- `corepack yarn eslint src/components/settings/api-keys-section.tsx`
- `corepack yarn typecheck`
- `npm run version:check`
- `git diff --check`
- `git diff --cached --check`
- `gitleaks protect --staged --no-banner --redact`

## Notes

`corepack yarn lint` is currently blocked by existing unrelated lint errors outside this change:

- `web/src/app/[locale]/invite/[token]/page.tsx:24`
- `web/src/components/layout/sidebar.tsx:33`

I did not run the dashboard manually because full local app startup requires environment-backed services that are outside this fix.

### DIFF
--- a/web/src/components/settings/api-keys-section.tsx
+++ b/web/src/components/settings/api-keys-section.tsx
@@ -34,6 +34,15 @@ interface ApiKey {
   created_at: string;
 }
 
+const configuredAppUrl = process.env.NEXT_PUBLIC_APP_URL?.replace(/\/$/, "") ?? "";
+
+function getMcpEndpoint(): string {
+  const appUrl =
+    configuredAppUrl ||
+    (typeof window !== "undefined" ? window.location.origin : "");
+  return `${appUrl}/api/mcp`;
+}
+
 function formatDate(iso: string | null): string {
   if (!iso) return "—";
   return new Date(iso).toLocaleDateString(undefined, {
@@ -50,6 +59,7 @@ export function ApiKeysSection() {
   const [newName, setNewName] = useState("");
   const [creating, setCreating] = useState(false);
   const [revealedToken, setRevealedToken] = useState<string | null>(null);
+  const mcpEndpoint = getMcpEndpoint();
 
   const load = useCallback(async () => {
     setLoading(true);
@@ -141,16 +151,13 @@ export function ApiKeysSection() {
           <p className="font-medium text-foreground">MCP endpoint</p>
           <div className="flex items-center gap-2">
             <code className="flex-1 truncate font-mono">
-              {typeof window !== "undefined" ? window.location.origin : ""}
-              /api/mcp
+              {mcpEndpoint}
             </code>
             <Button
               size="sm"
               variant="ghost"
               className="h-6 px-2"
-              onClick={() =>
-                copy(`${window.location.origin}/api/mcp`)
-              }
+              onClick={() => copy(mcpEndpoint)}
             >
               <Copy className="h-3 w-3" />
             </Button>


gkhngyk:
Welcome to Ansvisor, and thank you for the contribution! Merging now.

Nice clean fix — the module-level `configuredAppUrl` constant + trailing-slash normalize were both bonus moves the issue didn't ask for and both are the right defensive choices. Collapsing the two call sites into a single `mcpEndpoint` variable is also a tidier shape than what the issue described.

If you'd like to keep an eye on where Ansvisor goes next, a ⭐ on the repo helps others discover it (and Watch sends you a ping when related issues land). Either way — looking forward to your next one.

PR #48, opened 2026-05-18
fix(citations): group raw model slugs under display names in the Platforms filter

sourabh-31:
- What changed and why

This updates the /dashboard/citations Platforms filter to match the behavior already used on the Answer Engine Insights page.

Before this change, Citations showed raw model slugs like gpt-5-3-mini and gpt-5-5 as separate dropdown options. That was confusing for users, who expect a single ChatGPT option representing the whole model family.

This PR fixes that in two places:

Frontend: the Citations page now groups observed model slugs by their resolved display label and builds dropdown options as { value, label }, where value is a sorted comma-joined slug list and label is the user-facing family name like ChatGPT.
Backend: the Citations query now supports comma-joined model values and applies them against model_used, so selecting one grouped option matches all underlying model slugs in that family.
This brings Citations in line with Insights and ensures platform filters are consistent across the dashboard.

- How to test the change

1. Open /dashboard/citations for a brand that has tracking data across multiple versions of the same AI family, such as multiple ChatGPT model slugs.
2. Open the Platforms dropdown.
3. Confirm that raw slugs like gpt-5-3-mini and gpt-5-5 no longer appear as separate options.
4. Confirm that a single ChatGPT option is shown instead.
5. Select ChatGPT and verify the citations list filters to rows from any of the underlying ChatGPT model slugs.
6. Repeat the same check for any other multi-version families available in the data, such as Claude, Gemini, Perplexity, or Copilot.
7. Confirm that single-version platforms still display a friendly label rather than the raw slug.
8. Clear the filter and verify the unfiltered citations list returns correctly.


gkhngyk:
Thanks for the work on this — the feature code itself (`page.tsx` + `citations.ts`, ~120 lines) follows the Insights page pattern cleanly and looks solid.

But there's a packaging issue blocking the merge: the diff includes ~16k lines of lock-file noise that need to be reverted before we can review / land this safely:

- `web/yarn.lock` — completely rewritten. The repo uses **Yarn Classic (v1)**; looks like your local setup is on **Yarn Berry**, which rewrote the lockfile into the Berry format. Please revert this file to `main`'s version.
- `web/.yarnrc.yml` — new file from Berry. This would force every other contributor onto Berry too, breaking their workspaces. Please delete it.
- `server/package-lock.json` — 5 unrelated lines drifted; revert to `main`'s version.

Easiest fix: on a fresh clone with Yarn Classic (`npm i -g yarn@1.22.22` if needed), check out your branch and reset just those files:

```bash
git checkout main -- web/yarn.lock server/package-lock.json
rm web/.yarnrc.yml
git add -u
git commit -m "chore: revert unrelated lockfile churn"
git push
```

Once the diff is just the citations changes (~120 lines across two files), we're good to go. The feature itself is exactly what the issue asked for.

gkhngyk:
Side note while you're cleaning up the lockfile churn — your accidental regeneration actually surfaced a real metadata bug we hadn't noticed: `server/package-lock.json` has been carrying stale `name` / `version` / `license` ("ISC" instead of the actual "MIT") since before the project was renamed.

I split it out as #50 so it doesn't get tangled with this PR. If you'd like to send a one-commit follow-up PR fixing that (literally `cd server && npm install`), it's yours — credit where it's due for spotting it.

PR #52, opened 2026-05-18
feat(sidebar): added user profile as (avatar + name) → settings

### LINKED ISSUE #45: feat(sidebar): show signed-in user profile row at the bottom (avatar + name → Settings)
## What

Add a user profile row at the bottom of the dashboard sidebar,
**above the existing collapse toggle**. The row shows:

- A circular avatar on the left with the user's **initials only**
  (no photo). Dark background, light text.
- The user's full name (or email if name is not set) to the right of
  the avatar.

Clicking the row navigates to `/dashboard/settings`.

When the sidebar is collapsed (icon-only mode), only the avatar
remains visible — same `isCollapsed` pattern the nav items use.

## Why

There's currently no quick visual indicator of *who's signed in*
inside the dashboard chrome — the only way to confirm your account
is to navigate to Settings. A persistent profile row in the sidebar:

- Gives instant identity confirmation (important on shared / team
  accounts)
- Adds a one-click shortcut to Settings without hunting in nav
- Matches the convention used by every dashboard product the user
  has seen (Notion, Linear, Vercel, Supabase, etc.)

## Where to add it

`web/src/components/layout/sidebar.tsx` — insert the new profile
block **immediately before** the existing `{/* Collapse toggle */}`
section (around line 170). Keep the collapse toggle exactly where
and how it is — the user row sits on top of it.

The two blocks together (profile row + collapse toggle) form the
sidebar's bottom stack, separated from the nav above by the existing
`ScrollArea` boundary.

## How to read the signed-in user

The auth store at `web/src/stores/use-auth-store.ts` already exposes
the current Supabase `user`. Read it via:

```ts
import { useAuthStore } from '@/stores/use-auth-store';
const user = useAuthStore((s) => s.user);
```

The user object's relevant fields:

- `user.user_metadata?.full_name` — preferred display name
- `user.email` — fallback when no name is set

If `user` is `null` (auth still loading), render nothing for the
profile row — don't show a placeholder.

## Initials logic

Compute from the name:

```ts
function getInitials(user: User): string {
  const name = user.user_metadata?.full_name?.trim();
  if (name) {
    const parts = name.split(/\s+/).filter(Boolean);
    const first = parts[0]?.[0] ?? '';
    const last = parts.length > 1 ? parts[parts.length - 1][0] : '';
    return (first + last).toUpperCase().slice(0, 2);
  }
  // No name → fall back to email's first letter
  return user.email?.[0]?.toUpperCase() ?? '?';
}
```

So `"Ada Lovelace"` → `AL`, `"Cher"` → `C`, no name + email
`ada@x.com` → `A`.

## Avatar styling

Use the existing shadcn primitives from
`@/components/ui/avatar` (the `Avatar` + `AvatarFallback`
components are already in the codebase). The `AvatarFallback` is
where the initials render.

For the dark background + light text requirement:

```tsx
<Avatar className="h-8 w-8 shrink-0">
  <AvatarFallback className="bg-foreground text-background text-xs font-semibold">
    {initials}
  </AvatarFallback>
</Avatar>
```

Using `bg-foreground` / `text-background` (Tailwind/shadcn tokens)
guarantees correct dark/light contrast in **both** themes
automatically — in light mode the avatar is near-black with white
text, in dark mode it's near-white with dark text. Don't hardcode
hex colours.

## The clickable row

The whole row should be a `Link` from `@/i18n/navigation` (already
imported at the top of the file) pointing at `/dashboard/settings`.
Style it the same way as the nav items above it — hover state, focus
ring, etc. Suggested shape:

```tsx
{!user ? null : (
  <div className="border-t p-2">
    <Link
      href="/dashboard/settings"
      className={cn(
        'flex items-center gap-3 rounded-md px-2 py-2 transition-colors',
        'hover:bg-accent hover:text-accent-foreground',
        isCollapsed && 'justify-center',
      )}
      title={isCollapsed ? displayName : undefined}
    >
      <Avatar className="h-8 w-8 shrink-0">
        <AvatarFallback className="bg-foreground text-background text-xs font-semibold">
          {initials}
        </AvatarFallback>
      </Avatar>
      {!isCollapsed && (
        <span className="truncate text-sm font-medium">
          {displayName}
        </span>
      )}
    </Link>
  </div>
)}
```

Where `displayName` is `user.user_metadata?.full_name ?? user.email`.

## How to verify

- [ ] Sign in → profile row appears at the bottom of the sidebar,
      above the collapse toggle
- [ ] Avatar shows correct initials: `"Ada Lovelace"` → `AL`,
      single-word name → first letter only, no-name user → first
      letter of email
- [ ] Click the row → navigates to `/dashboard/settings`
- [ ] Collapse the sidebar → only the avatar is visible, name is
      hidden, hover tooltip still shows the name
- [ ] Toggle dark mode → avatar background and text invert
      correctly (still high-contrast in both themes)
- [ ] Very long name (e.g. `"Maria Anne Theresa Johansson"`) →
      truncates cleanly with ellipsis, doesn't overflow the sidebar
- [ ] Signed-out state (e.g. brief moment during sign-out) →
      profile row is hidden, collapse toggle still works

## Out of scope

- A full dropdown menu on the profile row (sign out, switch account,
  etc.) — single click → Settings is enough for now
- Photo / avatar image upload — initials only, by design
- Showing the user's organisation / role — name only
- Editing the profile inline — clicking it just navigates

## Contribution guide

See [CONTRIBUTING.md](https://github.com/aeohub/ansvisor/blob/main/CONTRIBUTING.md) for branch naming, commit
conventions, and PR workflow.

sourabh-31:
### What Changed And Why

- Added a shared profile nav component in user-profile-nav-item.tsx and reused it in both sidebar.tsx and mobile-nav.tsx. This keeps the user profile UI consistent across desktop and mobile.
- Added a safe loading fallback in the shared profile item: while auth data is not ready, the avatar shows U and the label shows User Name, so the UI never looks empty.
- The profile label uses truncate so long names or emails do not break the sidebar/mobile-nav layout.
- Avatar initials are derived from only the first two name parts, so a name like Jack D Harlow would render as JD instead of using more than two initials.
- Updated mobile nav so the menu items are scrollable and the user profile stays fixed at the bottom.
- Fixed the collapsed desktop sidebar header so the logo is centered properly when minimized.

---

### How To Test The Change

1. Open the dashboard on desktop and verify the profile row appears at the bottom of the sidebar above the collapse button.
2. Refresh the page and confirm the temporary fallback state can show avatar U and label User Name before auth data is available.
3. Verify a long display name or email is visually truncated and does not overflow the sidebar or mobile nav row.
4. Verify initials use only the first two name parts. For example, a user with First Middle Last should show FM.
5. Collapse the desktop sidebar and confirm the logo is centered and the profile switches to avatar-only.
6. Open the mobile nav sheet and confirm the nav items scroll while the profile row stays fixed at the bottom.
7. Tap the profile row in mobile nav and confirm the sheet closes and navigation goes to /dashboard/settings.

---

### Screenshots
image
image

This should fix #45

### DIFF
--- a/web/src/components/layout/mobile-nav.tsx
+++ b/web/src/components/layout/mobile-nav.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useTranslations } from 'next-intl';
 import { usePathname, Link } from '@/i18n/navigation';
 import { dashboardNav } from '@/config/dashboard';
@@ -10,8 +10,11 @@ import { cn } from '@/lib/utils';
 import { Button } from '@/components/ui/button';
 import { Badge } from '@/components/ui/badge';
 import { Sheet, SheetContent, SheetTrigger } from '@/components/ui/sheet';
+import { UserProfileNavItem } from '@/components/layout/user-profile-nav-item';
 import { useBrandStore } from '@/stores/use-brand-store';
 import { Crown, Menu, MessageSquareText } from 'lucide-react';
+import Image from 'next/image';
+import { useTheme } from 'next-themes';
 
 export function MobileNav() {
   const [open, setOpen] = useState(false);
@@ -20,6 +23,12 @@ export function MobileNav() {
   const tBrands = useTranslations('brands');
   const { canUse, requiredPlanFor, isCloud } = useFeatureGate();
   const { activeBrandId } = useBrandStore();
+  const { resolvedTheme } = useTheme();
+  const [mounted, setMounted] = useState(false);
+  useEffect(() => setMounted(true), []);
+
+  const logoSrc =
+    mounted && resolvedTheme === 'dark' ? '/logo_dark.svg' : '/logo_light.svg';
 
   const getLabel = (title: string): string => {
     const map: Record<string, string> = {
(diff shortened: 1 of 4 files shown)


gkhngyk:
Great structural work — the shared `UserProfileNavItem` component, the initials cap at two name parts, and the mobile nav improvements (scrollable list + bottom-pinned profile + logo in header) are all clean judgment calls.

One small ask before we merge:

Please revert `web/src/components/ui/sheet.tsx`. The 33-line diff there is pure formatting churn (double quotes → single, added semicolons, trailing commas) on a shadcn primitive this PR doesn't logically touch — it's just adding noise to the review surface. One-liner:

```bash
git checkout main -- web/src/components/ui/sheet.tsx
```

Once that's done, we're good to go.

gkhngyk:
Clean revert — merging now. Thanks for the quick turnaround, looking forward to the next one.

Issue #41, opened 2026-05-18
feat(prompts): add CSV export button on the All Prompts tab


gkhngyk:
## What

Add a **Export CSV** button to the All Prompts tab on the Prompts
page (`/dashboard/prompts?tab=all`). Clicking it downloads the
prompt inventory — text, targeting (platforms / models / regions),
estimated AI demand, and 30-day visibility health — as a CSV file.

Third of the CSV export rollout (see #38 for Insights, #39 for
Topics). Same pattern, different dataset and a richer column set
because three data sources are joined on this page.

## Why

The All Prompts tab is where marketers see *everything they're
tracking on a single brand*. A CSV makes it easy to:
- Inventory the prompt set for a stakeholder review
- Spot inactive prompts, missing topics, or under-targeted
  platforms
- Cross-reference search demand vs. actual visibility in a
  spreadsheet pivot

## Where the button goes

The toolbar that sits alongside the tabs lives at:

- `web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx`
  around line 450 — the `<div className="flex items-center justify-between gap-4 flex-wrap">` wrapping `TabsList` and the **Manage prompts** link.

Place the Export button **immediately before** the **Manage prompts**
link, in the same row, **only when the All Prompts tab is active**:

```tsx
{tab === 'all' && (
  <Button variant="outline" size="sm" onClick={handleExport} ...>
    <Download className="h-4 w-4" />
    Export CSV
  </Button>
)}
<Link href={...}>Manage prompts</Link>
```

Reason: on the **Insights** tab the data shape is different — an
Export button there would either confuse users or need a separate
export flow. Keep it scoped.

- Use `<Button>` from `@/components/ui/button`, `variant="outline"`,
  `size="sm"`
- Icon: `Download` from `lucide-react`

## What to export

The page already loads three related data structures:

1. `allPrompts: Prompt[]` — the prompt inventory (line 215)
2. `volumeByPromptId: Map<string, PromptVolume>` — optional estimated
   demand per prompt (line 419)
3. `visibility: Record<string, PromptVisibilitySummary>` — optional
   last-30-day health per prompt

The CSV joins all three by `prompt.id`. Columns (in this order):

| Column | Source |
| --- | --- |
| `text` | `p.text` |
| `topic_id` | `p.topicId ?? ""` |
| `category` | `p.category ?? ""` |
| `platforms` | `p.platforms.join(" | ")` |
[...]

PR #54, opened 2026-05-19
feat(prompts): added csv export button in all prompts tab

sourabh-31:
### What Changed and Why

- Added CSV export support to the Prompts dashboard.
- The export action was added only to the All Prompts tab, not the Insights tab, because the CSV output is based on the prompt list and its related table data.
- Reused the shared CSV serializer in csv.ts (line 9) to keep export formatting consistent.
- Kept the export handler inside prompts/page.tsx (line 452) since the exported columns, row mapping, and filename are specific to the Prompts page.
- Disabled the export action when there is no prompt data to avoid empty downloads.

### How to Test the Change

1. Open the Prompts dashboard for a brand with prompt data.
2. Go to the All Prompts tab and confirm the Export CSV action is visible and enabled.
3. Click Export CSV and verify a file downloads with a name like ansvisor_<brand-slug>_prompts_<yyyy-mm-dd>.csv.
4. Open the downloaded file and confirm the headers and rows match the data shown in the All Prompts table.
5. Switch to the Insights tab and confirm the CSV export action is not shown there.
6. Check a brand with no prompts and confirm the export action is disabled or unavailable as expected.
7. Spot-check rows with empty optional fields or comma-separated values to confirm the CSV formatting is valid.

### Screenshots
image

This closes issue #41

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
@@ -31,6 +31,7 @@ import {
   RefreshCw,
   Pencil,
   Settings2,
+  Download,
 } from 'lucide-react';
 import { cn } from '@/lib/utils';
 import { useBrandStore } from '@/stores/use-brand-store';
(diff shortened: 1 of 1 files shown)


gkhngyk:
Thank you @sourabh-31  🚀

PR #63, opened 2026-05-20
fix: preserve pending content opportunities

### LINKED ISSUE #62: fix(content): stop deleting fresh content opportunities on every regeneration
## What's wrong

Every time the tracking worker runs the content opportunity generator,
it **deletes all existing `status='new'` opportunities** for the brand
and inserts the LLM's fresh batch in their place. The LLM tends to
produce ~10 suggestions per run, and those 10 are usually 80–90% the
same recommendations as yesterday's (because the underlying
visibility / competitor data barely changed overnight).

End result for the user:
- Day 1: clicks Content Opportunities → sees 10 suggestions, picks
  one to work on, leaves the rest open
- Day 2: opens the page → yesterday's open suggestions are **gone**,
  10 nearly-identical new ones in their place

Effectively the page loses the user's mental "to-do" state every
day. The `new` status is treated as ephemeral when it should be the
"untouched by the user yet" state.

## Steps to reproduce

1. Run a tracking job for a brand → confirm `content_opportunities`
   has ~10 rows with `status='new'`
2. Wait for the next tracking run (or trigger another)
3. Inspect `content_opportunities` again → the row IDs have all
   changed, the previous `new` rows are gone

## Where to fix

`server/src/lib/opportunity-generator.js` — around lines 143–147:

```js
await supabaseAdmin
  .from('content_opportunities')
  .delete()
  .eq('brand_id', brandId)
  .in('status', ['new']);
```

This delete block is the bug. Remove it, and instead **dedupe** the
LLM's batch against existing `new` rows so duplicates don't pile up.

## Suggested approach

```js
// Don't delete — accumulate. Fetch existing new opportunities and
// skip any LLM-proposed ones that already exist (same prompt + title).
const { data: existing } = await supabaseAdmin
  .from('content_opportunities')
  .select('prompt_id, title')
  .eq('brand_id', brandId)
  .eq('status', 'new');

const seen = new Set(
  (existing ?? []).map(
    (e) => `${e.prompt_id}::${(e.title ?? '').toLowerCase().trim()}`,
  ),
);

const freshRows = object.opportunities
  .map((opp) => {
    const rel = scored[opp.relatedPromptIndex] || scored[0];
    return { rel, opp };
  })
  .filter(({ rel, opp }) => {
    const key = `${rel.promptId}::${opp.title.toLowerCase().trim()}`;
    return !seen.has(key);
  })
  .map(({ rel, opp }) => ({
    brand_id: brandId,
    prompt_id: rel.promptId,
    title: opp.title,
    description: opp.description,
    type: opp.type,
    impact: opp.impact,
    opportunity_score: rel.score,
    status: 'new',
    source_data: {
      promptText: rel.text,
      estAiVolume: rel.estAiVolume,
      visibilityScore: rel.avgVisibility,
      competitorGap: rel.competitorGap,
      intent: rel.intent,
      keywords: rel.keywords,
      competitorsCited: rel.competitorsCited,
    },
  }));

if (freshRows.length > 0) {
  await supabaseAdmin.from('content_opportunities').insert(freshRows);
}
console.log(
  `[opportunities] Inserted ${freshRows.length} new opportunities for brand ${brandId} (${(existing ?? []).length} already pending)`,
);
```

Dedupe key is `${prompt_id}::${normalized_title}`. The LLM often
re-phrases the same suggestion slightly, but exact-title duplicates
are the most common case. We can refine the dedupe heuristic later
if needed; perfect dedupe is not required for this PR.

Statuses other than `new` (`sent`, `in_progress`, `done`, `dismissed`)
are already untouched by the old code — keep that behaviour.

## How to verify

- [ ] Trigger a tracking job for a brand → `content_opportunities`
      gets ~10 rows with `status='new'` (first run, nothing to
      accumulate against)
- [ ] Trigger another tracking job for the same brand → the original
      rows are **still there**, plus any genuinely new suggestions
      from the second run are appended (could be 0 if the LLM
      produced the same set, could be 1–3 if a few changed)
- [ ] Mark one of the new ones as `dismissed` from the UI → next
      tracking run does not "resurrect" it (dismiss already excludes
      it from the dedupe set since it's `status='dismissed'`, not
      `new` — so the same title could be inserted again; that's fine
      for a first pass, can be tightened later)
- [ ] Read the worker log → reports both how many were inserted and
      how many were already pending

## Out of scope

- Smarter dedupe (semantic similarity rather than exact-title match)
- Capping the total number of `new` opportunities to prevent runaway
  growth (in practice, LLM repetition keeps growth flat; revisit if
  someone hits 100+)
- Adding a "since last run" indicator in the UI to mark freshly
  inserted opportunities — separate UX issue

## Contribution guide

See [CONTRIBUTING.md](https://github.com/aeohub/ansvisor/blob/main/CONTRIBUTING.md) for branch
naming, commit conventions, and PR workflow.

ryonakae:
## Summary
- Stop deleting pending `new` content opportunities before each generation
- Dedupe generated opportunities against existing pending rows by `prompt_id` + normalized title
- Skip insert when every generated opportunity is already pending, and log inserted vs pending counts

Closes #62

## Verification
- `node --check server/src/lib/opportunity-generator.js`
- `git diff --check`

### DIFF
--- a/server/src/lib/opportunity-generator.js
+++ b/server/src/lib/opportunity-generator.js
@@ -140,15 +140,30 @@ Generate actionable content opportunities.`;
     prompt: userPrompt,
   });
 
-  await supabaseAdmin
+  const { data: existing } = await supabaseAdmin
     .from('content_opportunities')
-    .delete()
+    .select('prompt_id, title')
     .eq('brand_id', brandId)
-    .in('status', ['new']);
+    .eq('status', 'new');
 
-  const rows = object.opportunities.map((opp) => {
-    const rel = scored[opp.relatedPromptIndex] || scored[0];
-    return {
+  const seen = new Set(
+    (existing || []).map((opportunity) => (
+      `${opportunity.prompt_id}::${(opportunity.title || '').toLowerCase().trim()}`
+    )),
+  );
+
+  const rows = object.opportunities
+    .map((opp) => {
+      const rel = scored[opp.relatedPromptIndex] || scored[0];
+      return { rel, opp };
+    })
+    .filter(({ rel, opp }) => {
+      const key = `${rel.promptId}::${opp.title.toLowerCase().trim()}`;
+      if (seen.has(key)) return false;
+      seen.add(key);
+      return true;
+    })
+    .map(({ rel, opp }) => ({
       brand_id: brandId,
       prompt_id: rel.promptId,
       title: opp.title,
@@ -162,9 +177,10 @@ Generate actionable content opportunities.`;
         visibilityScore: rel.avgVisibility, competitorGap: rel.competitorGap,
         intent: rel.intent, keywords: rel.keywords, competitorsCited: rel.competitorsCited,
       },
-    };
-  });
+    }));
 
-  await supabaseAdmin.from('content_opportunities').insert(rows);
-  console.log(`[opportunities] Generated ${rows.length} opportunities for brand ${brandId}`);
+  if (rows.length > 0) await supabaseAdmin.from('content_opportunities').insert(rows);
+  console.log(
+    `[opportunities] Generated ${rows.length} new opportunities for brand ${brandId} (${(existing || []).length} already pending)`,
+  );
 }


gkhngyk:
Clean fix — exactly what the issue asked for. Merging shortly.

Nice catch on the intra-batch dedupe inside the filter (`seen.add(key)` after the lookup). The issue didn't call for it, but the LLM does occasionally rephrase the same title twice in one run, and that one line keeps the table tidy without any extra round-trips.

Welcome to Ansvisor — looking forward to the next one.

Issue #58, opened 2026-05-20
fix(onboarding): add sign out button to onboarding header


sourabh-31:
### What's wrong
---
When a new user signs up, they are redirected to the onboarding flow at `/dashboard/onboarding`.

If the user leaves onboarding midway without completing it, then later returns and tries to sign in with a different account, they currently cannot do so because the onboarding screen has no way to sign out.

This creates a dead-end state where the partially onboarded account remains active and the user is stuck inside onboarding unless they manually clear session data or use another browser.

### Steps to reproduce
---
1. Sign up with a new account
2. Get redirected to `/dashboard/onboarding`
3. Leave onboarding midway without completing it
4. Close the tab/browser
5. Return later and open the app again
6. Try to switch accounts or sign in with another account
7. Observe:
The user is automatically redirected back into onboarding for the existing session, but there is no visible sign-out action anywhere in the onboarding UI.

### Expected
---
Users should be able to sign out during onboarding so they can:
* switch accounts
* restart onboarding with another user
* recover from accidental or incomplete onboarding sessions

### Where to fix
---
Onboarding route:
* `app/[locale]/(onboarding)/dashboard/onboarding`

Relevant component:
* `marketing-header.tsx`

### Suggested fix
---
Reuse the existing `marketing-header.tsx` component pattern from the auth pages and add it to the onboarding flow.

Suggested header layout:
* Brand logo / wordmark on the left
* Sign out button on the right

Using the same flex + `justify-between` structure as the auth pages would keep the onboarding experience visually consistent with the rest of the authentication flow.

### Hint
---
The auth pages already use `marketing-header.tsx`, so the onboarding pages can likely reuse the same component structure with minimal additional UI work.

The sign out button should use the existing auth logout flow instead of implementing separate session-clearing logic.

### How to verify
---
* Start onboarding with Account A
* Leave onboarding incomplete
* Return to the app
* Confirm a visible Sign out button exists on `/dashboard/onboarding`
* Click Sign out
* Verify the session is cleared
* Sign in successfully with Account B
* Confirm onboarding works normally for the new account

### Contribution guide
---
See [CONTRIBUTING.md](https://github.com/aeohub/ansvisor/blob/main/CONTRIBUTING.md) for branch naming, commit conventions, and PR workflow.

gkhngyk:
## Design proposal: fixed bottom-right sign-out, not a full header

[...]

PR #68, opened 2026-05-20
fix(onboarding): added signout button in the bottom right corner

sourabh-31:
### What Changed And Why
---

- Added a new shared client component: onboarding-sign-out-button.tsx.
It renders a fixed bottom-right Sign out button with the LogOut icon, muted styling, hover treatment, focus ring, and a disabled state while sign-out is in progress.

- Mounted that button once in the onboarding layout: layout.tsx.
This makes it appear across the whole onboarding flow without repeating the same markup in every step.

- The sign-out handler uses Supabase auth.signOut() and then does window.location.href = '/sign-in'.
The reason for the full-page navigation is to force the server-rendered auth layout to read the cleared session immediately, instead of relying on a client-side route transition.

- This approach avoids reusing a full marketing header.
It keeps sign-out as a quiet escape hatch rather than giving it top-level visual priority or duplicating branding already present in onboarding.

### How To Test The Change
---

1. Sign in and enter onboarding. Confirm Sign out appears in the bottom-right on every onboarding step.
2. Resize the viewport to a short height, around 600px, and scroll longer steps like topics/prompts. Confirm the button stays fixed and visible.
3. Use Tab until the button is focused. Confirm the focus ring appears and Enter/Space activates it.
4. Click Sign out. Confirm you land on /sign-in and the session is cleared.
5. Try opening onboarding or dashboard again after signing out. Confirm you are no longer authenticated.
6. Sign in with a different account and confirm onboarding starts fresh.
7. Toggle dark mode and verify the muted text, hover background, and focus styles still look correct.

### Screenshot
---
image

This closes issue #58

### DIFF
--- a/web/messages/en.json
+++ b/web/messages/en.json
@@ -58,7 +58,8 @@
       "passwordRequired": "Password is required.",
       "passwordTooShort": "Password must be at least 8 characters.",
       "generic": "An error occurred. Please try again.",
-      "emailNotConfirmed": "Please verify your email address first. Check your inbox for the confirmation link."
+      "emailNotConfirmed": "Please verify your email address first. Check your inbox for the confirmation link.",
+      "signOutError": "Failed to sign out. Please try again."
     }
   },
   "dashboard": {
--- a/web/src/app/[locale]/(onboarding)/layout.tsx
+++ b/web/src/app/[locale]/(onboarding)/layout.tsx
@@ -1,7 +1,7 @@
-import { redirect } from 'next/navigation';
-import { createClient } from '@/lib/supabase/server';
-import { AuthProvider } from '@/components/providers/auth-provider';
-import { OnboardingSignOutButton } from '@/components/auth/onboarding-sign-out-button';
+import { redirect } from "next/navigation";
+import { createClient } from "@/lib/supabase/server";
+import { AuthProvider } from "@/components/providers/auth-provider";
+import { OnboardingSignOutButton } from "@/components/auth/onboarding-sign-out-button";
 
 export default async function OnboardingLayout({
   children,
@@ -14,7 +14,7 @@ export default async function OnboardingLayout({
   } = await supabase.auth.getUser();
 
   if (!user) {
-    redirect('/sign-in');
+    redirect("/sign-in");
   }
 
   return (
--- a/web/src/app/[locale]/(onboarding)/layout.tsx
+++ b/web/src/app/[locale]/(onboarding)/layout.tsx
@@ -1,6 +1,7 @@
-import { redirect } from "next/navigation";
-import { createClient } from "@/lib/supabase/server";
-import { AuthProvider } from "@/components/providers/auth-provider";
+import { redirect } from 'next/navigation';
+import { createClient } from '@/lib/supabase/server';
+import { AuthProvider } from '@/components/providers/auth-provider';
+import { OnboardingSignOutButton } from '@/components/auth/onboarding-sign-out-button';
 
 export default async function OnboardingLayout({
   children,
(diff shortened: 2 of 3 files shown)


gkhngyk (changes requested):
Thanks for picking this one up, Sourabh — the component design is clean and mounting it once in the layout is the right call.

One critical thing to fix before this can merge, plus two small nits.

### Critical: existing placeholder is still rendered (overlap)

There's already a placeholder `SignOutLink` in `web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx:72` that's rendered in six step views (lines 997, 1090, 1177, 1346, 1428, 1586). It uses the exact same `fixed bottom-6 right-6` position and styling as your new button — so as it stands the PR would render **two stacked Sign out buttons** on every onboarding step.

Fix: remove all six `<SignOutLink />` usages in `page.tsx` and the function definition at line 72 (it'll be orphaned once the call sites go). That makes the new layout-level button the single source of truth — which was the whole point of mounting it in the layout.

### Nits (optional)

1. **`layout.tsx` quote-style churn** — the diff reformats three existing import lines from double quotes to single quotes. The rest of `web/src/components/auth/` uses double quotes, so reverting those three lines would shrink the diff to just the meaningful change and keeps the file consistent with its neighbors.

2. **Hardcoded English toast** — the button label uses `tAuth('signOut')` but the error toast is a fixed English string. Not a blocker since sign-out errors are rare, but an `auth.signOutError` key would match the rest of the component.

Let me know once the overlap is resolved and I'll re-review. Thanks again!

sourabh-31:
> Thanks for picking this one up, Sourabh — the component design is clean and mounting it once in the layout is the right call.
> 
> One critical thing to fix before this can merge, plus two small nits.
> 
> ### Critical: existing placeholder is still rendered (overlap)
> There's already a placeholder `SignOutLink` in `web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx:72` that's rendered in six step views (lines 997, 1090, 1177, 1346, 1428, 1586). It uses the exact same `fixed bottom-6 right-6` position and styling as your new button — so as it stands the PR would render **two stacked Sign out buttons** on every onboarding step.
> 
> Fix: remove all six `<SignOutLink />` usages in `page.tsx` and the function definition at line 72 (it'll be orphaned once the call sites go). That makes the new layout-level button the single source of truth — which was the whole point of mounting it in the layout.
> 
> ### Nits (optional)
> 1. **`layout.tsx` quote-style churn** — the diff reformats three existing import lines from double quotes to single quotes. The rest of `web/src/components/auth/` uses double quotes, so reverting those three lines would shrink the diff to just the meaningful change and keeps the file consistent with its neighbors.
> 2. **Hardcoded English toast** — the button label uses `tAuth('signOut')` but the error toast is a fixed English string. Not a blocker since sign-out errors are rare, but an `auth.signOutError` key would match the rest of the component.
> 
> Let me know once the overlap is resolved and I'll re-review. Thanks again!

Hey, thanks for the detailed review — nits are addressed!

On the critical item: I looked through `web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx` but couldn't find the `SignOutLink` definition at line 72 or any of the six call sites you mentioned. It's possible the file has shifted since your review, or it may have been removed in a prior commit.

Could you double-check? If you're still seeing it on your end, a quick snippet or the exact line range would help me track it down. Happy to clean it up if it's there — just want to make sure I'm not deleting the wrong thing.

gkhngyk:
You're right — I went back and checked, the file on `main` doesn't have that placeholder. No overlap, nothing else to fix.

Both nits are addressed cleanly. Approving and merging. Thanks for the careful work!

Issue #93, opened 2026-05-25
perf(insights): speed up the dashboard insights page (DB-side aggregation + cache with invalidation)


gkhngyk:
## Problem

`/dashboard/insights` is slow to load. The cause is the data layer, not the UI:

- The insights server actions in [`web/src/lib/actions/tracking.ts`](https://github.com/aeohub/ansvisor/blob/main/web/src/lib/actions/tracking.ts) pull **full rows** (`from('prompt_results').select('*')`) and **aggregate in JS** — `getInsightsSummary`, `getShareOfVoiceData`, and `getCompetitorComparison` each loop over the fetched rows (`reduce`/`for…of`) instead of letting Postgres compute the totals. As a brand accumulates results (thousands of rows), this transfers a lot of data and computes in Node on every load.
- The page ([`insights/page.tsx`](https://github.com/aeohub/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/page.tsx)) fires **4–5 of these actions in parallel** on every mount, each doing multiple Supabase round-trips (prompt_results + prompts + topics).
- **No caching** anywhere on this path.
- `prompt_results` has single-column indexes (`brand_id`, `created_at`, `prompt_id`) but **no composite `(brand_id, created_at)`** — the exact shape the brand + date-range filters need.

## Goal

Make the insights page fast **without** hiding fresh data: on-demand **Run All / Refresh results / Test Single Prompt** must still reflect immediately (these update `prompt_results` synchronously). So a blind long-TTL cache is a non-goal — freshness on write is a hard requirement.

## Proposed approach (phased)

**1. Aggregate in the DB instead of JS (biggest win, cache-independent)**
- Replace `select('*')` + JS reduce with SQL aggregation — Postgres RPC functions or views that return the summary / share-of-voice / competitor metrics already computed. Select only needed columns.
- Add a composite index `(brand_id, created_at DESC)` on `prompt_results` (and any others the new queries need).

**2. (Optional, larger) Precomputed daily summaries**
- Scheduled tracking runs ~daily. Have the tracking worker write a per-brand/day summary row (e.g. `daily_insights`) on each run, so reads become trivial lookups that are always fresh (refreshed exactly when data changes — no TTL/staleness tradeoff).

**3. Caching with invalidation (complement)**
[...]

PR #114, opened 2026-05-28
perf(insights): aggregate insights data in Postgres instead of pulling rows into Node

gkhngyk:
Implements phase 1 of #93. The largest, no-tradeoff lever from the issue: move the three insights server-action reducers out of Node and into Postgres without introducing any TTL-bound cache that would delay on-demand refresh updates.

## The shape of the bug

\`getInsightsSummary\`, \`getShareOfVoiceData\`, and \`getCompetitorComparison\` all followed the same pattern: \`select('*')\` from \`prompt_results\` for the brand + filter set, then \`reduce\`/\`Map\`/\`for…of\` the rows in JS. Each row carries the **full AI response text** (kilobytes), \`citations\` jsonb, and \`competitor_mentions\` jsonb. For a brand with thousands of results that's a multi-hundred-MB transfer + GC-heavy reducers on every insights page load — and the page fires 4–5 of these actions in parallel on mount.

### 1. Three new RPCs

In a single new migration \`supabase/migrations/00006_insights_aggregates.sql\`:

| Function | Replaces | Returns |
|---|---|---|
| \`insights_aggregates\` | \`getInsightsSummary\`'s reducer | Totals, sums, last_checked_at, per-model breakdown |
| \`competitor_aggregates\` | \`getCompetitorComparison\`'s reducer | Brand + per-competitor sums, per-provider grids for the table |
| \`share_of_voice_aggregates\` | \`getShareOfVoiceData\`'s reducer | Totals, per (model_used, platform), per-day buckets |

Each returns a single JSONB row of raw **sums + counts**, never pre-divided averages. The final \`/\` + \`Math.round\` still happens in JS exactly as before — guarantees the displayed numbers don't drift by even ±1 from a half-rule mismatch between JS's \`Math.round\` and Postgres \`ROUND\`. The provider mapping (\`resolveProvider\`) also stays in JS so the SQL doesn't ship a duplicate lookup that has to be kept in sync as engines land.

\`SECURITY DEFINER\` to match the existing \`get_latest_prompt_results\` pattern; route layer still verifies brand-belongs-to-org access before calling.

### 2. Composite index

\`CREATE INDEX (brand_id, created_at DESC) ON prompt_results\` — the exact shape every insights filter needs. Existing indexes are single-column on \`brand_id\` and \`created_at\` separately, forcing the planner to pick one and re-sort/filter on the other in the heap. The composite serves both the old query path (still used by listing endpoints like \`getPromptResults\`, \`getVisibilityTrend\`) and the new RPC \`WHERE\`s.

### 3. Refactor of three action functions

\`web/src/lib/actions/tracking.ts\` — the three problem functions now do:
- One RPC call for the displayed window
- Two parallel RPC calls (current 7-day vs previous 7-day) for the delta
- A trivial fold over the small grouped arrays the RPC returns
- \`resolveProvider\` over those grouped rows for provider buckets

\`buildResultsQuery\` and \`applyModelFilter\` stay — they're still used by row-listing endpoints (\`getPromptResults\`, \`getVisibilityTrend\`).

## Parity verification

Before merging anything, ran the RPCs against a real brand with **5054 prompt_results** in prod and diffed every field against the equivalent inline-SQL aggregation that mirrors the JS reducer arithmetic. Every field matched byte-for-byte at decimal precision (e.g. \`sum_visibility = 384907.7615259740259744489904\` from both paths). Verified for:

- \`insights_aggregates\` — unfiltered, plus with a 30-day date filter
- \`competitor_aggregates\` — brand_total_mentions, brand_total_citations, sum-of-competitor-mentions, provider combo counts, competitor × provider combo counts (70 distinct combos checked)
- \`share_of_voice_aggregates\` — totals, distinct day count, distinct (model_used, platform) combos

Local smoke load of \`/dashboard/insights\` then renders the same KPIs, breakdowns, and trend lines with the new path — no console errors, no server-side RPC errors, filters work.

## Why this is the safe rewrite

- **Pure addition** at the schema level: new index, new functions. No table altered, no column changed, no row touched. Drop the functions and the index → table is identical.
- **Read-side only**: tracking worker writes, on-demand Run All / Refresh / Test Single Prompt writes — none of these change. \`prompt_results.shopping_cards\` capture and every other write path stays exactly as-is.
- **No TTL cache introduced**: explicit constraint in the issue (Run All / Refresh must reflect immediately). This PR achieves the speedup *without* hiding fresh data behind any cache window.
- **JS-side rounding preserved**: the only practical drift risk on a SQL refactor like this is rounding/half-rule mismatch. By returning raw sums + counts and doing the final divide + round in JS, that risk class is closed.
- **Reversible**: \`git revert\` on the code commit alone returns behavior to the old \`select('*')\` path; the migration's functions and index sit idle if unused.

## Files

| File | Change |
|---|---|
| \`supabase/migrations/00006_insights_aggregates.sql\` | New migration: composite index + three RPCs |
| \`web/src/lib/actions/tracking.ts\` | Three insights actions refactored to call the RPCs; new \`modelFilterArray\` helper + RPC return-shape interfaces |
| \`web/src/types/supabase.ts\` | Added \`Functions\` entries for the three new RPCs so the \`.rpc()\` call sites type-check |

## Deploy order

The migration has already been applied to the cloud project (additive, reversible). Self-hosters pick it up at the next release via \`supabase db push\` — standard prosedür per CONTRIBUTING.

## Out of scope (deferred, from #93)

- **Precomputed daily summary table** (phase 2) — only worth it if phase 1 doesn't move the needle enough; measure first.
- **Use cache / cacheTag invalidation on insights reads** (phase 3) — complement to this PR, layers on top of these RPCs.

### DIFF
--- a/supabase/migrations/00006_insights_aggregates.sql
+++ b/supabase/migrations/00006_insights_aggregates.sql
@@ -1,0 +1,347 @@
+-- Phase-1 perf fix for /dashboard/insights (#93)
+--
+-- Today the insights server actions in web/src/lib/actions/tracking.ts pull
+-- `SELECT *` from prompt_results and aggregate in JS. Each row carries the
+-- full AI response text (kilobytes) + JSONB citations/competitor_mentions,
+-- and for a brand with thousands of results the page transfers hundreds of
+-- MB on every load + does big reducers in Node memory.
+--
+-- This migration:
+--   1. Adds a composite (brand_id, created_at DESC) index — the exact shape
+--      every insights filter needs but the only existing indexes are
+--      single-column.
+--   2. Adds three RPC functions that perform the aggregation server-side and
+--      return only the totals + small grouped slices. Callers can compute
+--      the same final numbers from these outputs with bit-for-bit parity
+--      against the existing JS reducers (parity test in
+--      scripts/parity-check-insights.ts).
+--
+-- Functions intentionally return RAW SUMS + COUNTS (not pre-divided averages)
+-- so the final round happens in JS exactly as it does today — guarantees the
+-- displayed dashboard numbers don't drift by even ±1 from a `Math.round`
+-- half-rule mismatch between JS and Postgres.
+--
+-- Security model matches the existing get_latest_prompt_results functions:
+-- SECURITY DEFINER — server actions verify brand-belongs-to-org access at
+-- the route layer before calling. Adding a defensive org check inside the
+-- function is a separate hygiene task tracked elsewhere.
+
+-- ── Index ─────────────────────────────────────────────────────────────────
+-- Composite covers WHERE brand_id = ? AND created_at BETWEEN ? AND ?
+-- ORDER BY created_at DESC — single seek + sequential index walk instead of
+-- index scan + heap filter + sort. ~11k rows in prod today; CREATE INDEX
+-- without CONCURRENTLY locks writes for milliseconds. CONCURRENTLY would
+-- avoid the lock but cannot run inside a transaction, and Supabase wraps
+-- migrations in one. Plain CREATE is the right call at this row count.
+CREATE INDEX IF NOT EXISTS idx_prompt_results_brand_created
+  ON public.prompt_results USING btree (brand_id, created_at DESC);
+
+
+-- ── insights_summary ──────────────────────────────────────────────────────
+-- Replaces getInsightsSummary's `select('*') + JS reduce` over the filtered
+-- row set. Returns one JSONB object with raw sums + counts + the per-model
+-- breakdown shape the page needs.
+CREATE OR REPLACE FUNCTION public.insights_aggregates(
+  p_brand_id   uuid,
+  p_platform   text         DEFAULT NULL,
+  p_models     text[]       DEFAULT NULL,
+  p_region     text         DEFAULT NULL,
+  p_date_from  timestamptz  DEFAULT NULL,
+  p_date_to    timestamptz  DEFAULT NULL,
+  p_prompt_id  uuid         DEFAULT NULL,
+  p_topic_id   uuid         DEFAULT NULL
+)
+RETURNS jsonb
+LANGUAGE sql
+STABLE
+SECURITY DEFINER
+SET search_path = public
+AS $$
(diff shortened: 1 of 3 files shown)


gkhngyk:
Thanks @copilot-pull-request-reviewer — useful pass, all six points are valid. Addressed five in $(git rev-parse --short HEAD) and tracked the sixth as a follow-up.

| # | Concern | Status |
|---|---|---|
| 1 | \`jsonb_agg(by_model)\` non-deterministic order → platform chart jitter | **Fixed** — added \`ORDER BY\` inside every \`jsonb_agg(by_*)\` across all three RPCs (by_model on result_count DESC, by_competitor on row_count DESC, provider grids on (platform, model_used) NULLS LAST, by_platform on SoV same). |
| 2 | \`insights_aggregates\` delta calls swallow RPC errors | **Fixed** — capture \`error\` from both current- and previous-window calls, throw on either. |
| 3 | \`competitor_aggregates\` delta calls swallow errors | **Fixed** — same pattern. |
| 4 | \`share_of_voice_aggregates\` previous-period call swallows error | **Fixed** — same pattern. |
| 5 | \`compByProvider\` keyed by \`competitor_name\` collides under \`''\` for unnamed competitors | **Fixed** — centralized a \`competitorDisplayName(name, id)\` helper that falls back to the competitor id when the name is empty, applied consistently to both the \`entries[]\` name field and the \`compByProvider\` key so the providerRows column header still matches the table row label. |
| 6 | \`SECURITY DEFINER\` on the RPCs allows cross-org reads with arbitrary \`p_brand_id\` | **Deferred — issue #115** | 

On #6 specifically: the concern is real and worth fixing, but the same property exists on the pre-existing \`get_latest_prompt_results\` RPCs from \`00001_initial_schema.sql\` — sweeping all of them with one consistent membership-check pattern (or moving them to \`SECURITY INVOKER\`) belongs in a hygiene PR rather than this perf-focused one. Tracking in #115.

Re-applied the migration to the cloud project via \`CREATE OR REPLACE\`; parity test against the same 5054-row brand still matches byte-for-byte after the ORDER BY additions.

PR #147, opened 2026-06-02
fix(web): show role label instead of raw enum in Team settings selects

### LINKED ISSUE #145: Settings → Team role select shows slug (`agency_partner`) instead of label (`Agency Partner`)
## Problem

In **Settings → Team**, the role `<Select>` shows the raw enum value (`agency_partner`) instead of the human label (`Agency Partner`) after a selection is made. Happens in two places:

- Existing-member role-change dropdown next to each row
- The role dropdown inside the **Invite member** dialog

Repro:
1. Open Settings → Team.
2. Click the role dropdown for any member, pick **Agency Partner**.
3. The trigger button collapses to show `agency_partner` instead of `Agency Partner`.

Same thing happens in the invite dialog when you pick **Agency Partner** for the new invite.

## Root cause

The project uses [Base UI](https://base-ui.com/) for select primitives ([`web/src/components/ui/select.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/ui/select.tsx)). Base UI's `Select.Value` defaults to rendering the raw selected `value` string when no children are passed — it does **not** pull the matched `SelectItem`'s children automatically (unlike Radix).

Both selects in [`web/src/components/settings/team-section.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/team-section.tsx) render `<SelectValue />` with no children, so the raw role enum (`admin`, `manager`, `analyst`, `agency_partner`) leaks into the UI for the only role whose slug differs from its label.

## Fix

Use Base UI's render-prop pattern on `SelectValue`, mapping the selected value to its label via the existing [`roleLabel`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/team-section.tsx#L70) helper that already lives in the file:

```tsx
<SelectValue placeholder="Select role">
  {(value) => (value ? roleLabel(value as TeamRole) : 'Select role')}
</SelectValue>
```

There's an in-repo reference for the same pattern at [`web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx:687-693`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx) — copy that shape.

Apply to both `<SelectValue />` instances in the file (line ~278 for the row-level role select, line ~507 for the invite dialog).

## Acceptance criteria

- [ ] Picking **Agency Partner** in either select shows **Agency Partner** in the trigger, not `agency_partner`
- [ ] Same for any other role (admin / manager / analyst — these already look right because their label equals their slug capitalized, but the fix should cover them uniformly)
- [ ] No visual regression on the row-level pending-update spinner — the trigger still swaps in `<Loader2>` while a role change is in flight

## Out of scope

- Any other slug-vs-label leak elsewhere in the app. If you find one while you're in there, leave a note and we'll open a separate issue.

akagifreeez:
## What & why

In **Settings -> Team**, the role `<Select>` displayed the raw enum value (e.g. `agency_partner`) instead of the human label (`Agency Partner`) after a selection — in both the existing-member role dropdown and the invite dialog.

Base UI's `<Select.Value />` renders the raw value unless the `items` map is provided on `<Select.Root>`. This passes the existing `ROLE_OPTIONS` array as `items` to both selects, so the human-readable label is shown.

Closes #145.

## How to test

1. Go to **Settings -> Team**.
2. Change an existing member's role via the dropdown — the control now shows e.g. **Agency Partner** instead of `agency_partner`.
3. Open the **Invite** dialog and pick a role — same: the label is shown.

## Notes

- Minimal change: reuses the existing `ROLE_OPTIONS` (single source of truth), adds no new strings, and leaves `SelectValue` untouched.
- Verified locally: `prettier --check`, `eslint`, and `tsc --noEmit` all pass.

### DIFF
--- a/web/src/components/settings/team-section.tsx
+++ b/web/src/components/settings/team-section.tsx
@@ -270,6 +270,7 @@ function MemberRow({
 
       {canEdit ? (
         <Select
+          items={ROLE_OPTIONS}
           value={member.role}
           onValueChange={(v) => handleRoleChange(v as TeamRole)}
           disabled={pendingRole !== null}
@@ -499,6 +500,7 @@ function InviteDialog({ canInvite, onInvited }: { canInvite: boolean; onInvited:
             <div className="space-y-2">
               <Label htmlFor="invite-role">Role</Label>
               <Select
+                items={ROLE_OPTIONS}
                 value={role}
                 onValueChange={(v) => setRole(v as TeamRole)}
                 disabled={submitting}


gkhngyk (approved):
Lovely fix, @akagifreeez — thanks for picking this up. The `items` prop on `<Select.Root>` is genuinely the cleanest path here. Verified in `node_modules/@base-ui/react/select/root/SelectRoot.d.ts:90` — the JSDoc explicitly says it makes `<Select.Value>` render the matched item's label instead of the raw value. Two-line diff, single source of truth on `ROLE_OPTIONS`, no new strings. Nice.

Approving — Vercel's preview-deploy gate fails for fork PRs (expected), but the format / lint / typecheck job will run once a maintainer clicks **Approve and run** on the Actions tab. Will merge once it's green.

Issue #149, opened 2026-06-02
Citations: add an "Own domain only" filter (currently we can only hide it, not isolate it)


gkhngyk:
## Problem

The Citations page lets users **hide** their own domain from the list (`Exclude own domain`) and show only **competitors** (`Competitors only`), but there's no way to isolate **just** their own domain — to answer the very common AEO question "which of my pages are AI engines actually citing?".

Today's two filters at [`web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx:586-610`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/citations/page.tsx):

| Filter | Behavior |
|---|---|
| `Exclude own domain` | Hides rows where category = `you` |
| `Competitors only` | Shows only rows where category = `competitor` |

`SOURCE_CATEGORIES` in [`web/src/lib/citations/classify.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/citations/classify.ts) already includes a `you` bucket — citations to the user's tracked domain — and the aggregation in [`web/src/lib/actions/citations.ts:241`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts) already discriminates on that category. The data is there; the UI just has no toggle that says "give me only my own".

## Proposed change

Add a third filter button **`Own domain only`** alongside the existing two. It mirrors the existing pattern:

- New field `ownOnly?: boolean` on `CitationsFilters` in [`web/src/lib/actions/citations.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts).
- In the aggregation loop, when `ownOnly === true`, skip rows where `category !== 'you'`.
- A new toggle button in the filter bar (around line 596) with the same `variant`/`size`/`className` shape as the other two.
- The three buttons are mutually exclusive: turning `Own domain only` on resets `excludeOwnDomain` and `competitorOnly` (they can't coexist semantically — "exclude own" while "show only own" is nonsensical). The same already happens between the existing two (line 604).

## Acceptance criteria

- [ ] New `Own domain only` button appears in the citations filter bar, sitting next to the existing two
- [ ] Toggling it ON filters the list (and the source-type/domain/URL panels) to only category = `you` rows
- [ ] Toggling it ON also turns OFF the other two filters when they're set, so the three behave mutually exclusively
[...]

Issue #146, opened 2026-06-02
chore(security): verify Cloro webhook signatures on /cloro/callback


gkhngyk:
## Problem

The `/cloro/callback` endpoint at [`server/src/server.js:233`](https://github.com/ansvisor/ansvisor/blob/main/server/src/server.js) accepts webhooks from Cloro with no signature verification:

```js
// Cloro webhook callback (public, no auth — security via task_id entropy + DB lookup)
app.post('/cloro/callback', async (req, res) => { … })
```

Today's defence is just "we look up `task.id` in `cloro_pending_tasks` and ignore unknown ids." That bounces opportunistic spam but doesn't stop a forged delivery where the attacker has obtained a valid pending `task_id` by any side channel (a leaked log line, a compromised Cloro account, network capture). A forged COMPLETED payload runs straight through `handleScraperResult`, inserting attacker-controlled response text, citations, mention counts, sentiment, and shopping cards into `prompt_results` for the matched brand — silently. Hard to detect after the fact because the row looks normal.

Cloro now publishes a signed-delivery option (HMAC-SHA256 over `timestamp.rawBody`, three headers: `X-Cloro-Signature`, `X-Cloro-Timestamp`, `X-Cloro-Webhook-Id`). Their docs are at https://docs.cloro.dev/integrations/webhooks. We should enable signing and verify on this endpoint.

## Scope

- Enable signing in the Cloro dashboard for the org; grab the generated `whsec_…` secret.
- Add `CLORO_WEBHOOK_SECRET` to Vercel env (production + preview + development).
- In [`server/src/server.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/server.js), mount `express.raw({ type: 'application/json' })` on `/cloro/callback` specifically so the body bytes survive intact. The default `express.json()` middleware re-serializes the payload and breaks HMAC.
- Compute `HMAC-SHA256(secret, `${timestamp}.${rawBody}`)`, compare against `X-Cloro-Signature` (`v1=…`) using `crypto.timingSafeEqual`. Reject 401 on mismatch.
- Reject 400 when `X-Cloro-Timestamp` is more than 5 minutes off `Date.now()` (replay protection).
- Parse the JSON body only **after** signature passes.
- Logging: bump a counter / log line when we reject so we can spot probing.

## Acceptance criteria

- [ ] Endpoint returns 401 for any POST without a valid signature
- [ ] Endpoint returns 400 for stale timestamps (> 5 min)
[...]

PR #150, opened 2026-06-02
fix(team): display role labels in selects

### LINKED ISSUE #145: Settings → Team role select shows slug (`agency_partner`) instead of label (`Agency Partner`)

adityabhatkar23:
## Summary

- render role labels instead of raw enum values in team role selects
- apply the fix to the member role dropdown
- apply the fix to the invite member dialog role dropdown

## Testing

- changed a member role to Agency Partner and verified the trigger displays "Agency Partner"
- selected Agency Partner in the invite dialog and verified the trigger displays "Agency Partner"
- verified the role update loading spinner still appears while updates are in flight

## Screenshot
Screenshot 2026-06-02 at 18-07-15 Ansvisor

Closes #145

### DIFF
--- a/web/src/components/settings/team-section.tsx
+++ b/web/src/components/settings/team-section.tsx
@@ -275,7 +275,13 @@ function MemberRow({
           disabled={pendingRole !== null}
         >
           <SelectTrigger size="sm" className="w-36">
-            {pendingRole ? <Loader2 className="h-3 w-3 animate-spin" /> : <SelectValue />}
+            {pendingRole ? (
+              <Loader2 className="h-3 w-3 animate-spin" />
+            ) : (
+              <SelectValue>
+                {(value) => (value ? roleLabel(value as TeamRole) : 'Select role')}
+              </SelectValue>
+            )}
           </SelectTrigger>
           <SelectContent>
             {ROLE_OPTIONS.map((opt) => (
@@ -504,7 +510,9 @@ function InviteDialog({ canInvite, onInvited }: { canInvite: boolean; onInvited:
                 disabled={submitting}
               >
                 <SelectTrigger className="w-full">
-                  <SelectValue />
+                  <SelectValue>
+                    {(value) => (value ? roleLabel(value as TeamRole) : 'Select role')}
+                  </SelectValue>
                 </SelectTrigger>
                 <SelectContent>
                   {ROLE_OPTIONS.map((opt) => (


gkhngyk:
Hey @adityabhatkar23 — thanks for following through on this, the work is genuinely good: render-prop pattern matches the in-repo reference (`insights/page.tsx`), explicit placeholder fallback, screenshot + test notes in the body, CI green. Textbook PR shape.

Unlucky timing though — #147 happened to land at 13:51 UTC (a few minutes before your CI cleared) so the issue is already closed by the same fix via a different route (`items` prop on `<Select.Root>` instead of a `SelectValue` render-prop — both are valid Base UI patterns). The work on #150 is sound, just no longer needed. Closing as superseded.

This race is on us, not on you — we don't have a strong "claim" mechanism beyond assigning, and #147 was opened without one. I unassigned you from #145 earlier and suggested some alternatives but our messages crossed.

If you want to pick up another one, three are sitting open right now:

- **#149** — Citations: add an "Own domain only" filter. **Same shape as #145** — you're adding a third toggle button next to two existing ones in the citations filter bar, plus one extra check in the aggregator. `good first issue` labeled.
- **#151** — Forgot Password flow (the link on `/sign-in` is dead). Bigger lift — two new routes, Supabase email template config, account-enumeration considerations — not `good first issue` but high impact if you're up for it.
- **#146** — Verify Cloro webhook signatures (Node.js HMAC verification). Security-sensitive so labeled `help wanted` rather than `good first issue`.

Comment on whichever and we'll assign you. Thanks again for the contribution — please don't let the race discourage you.

PR #164, opened 2026-06-03
feat: add "Own domain only" filter for citations isolation

### LINKED ISSUE #149: Citations: add an "Own domain only" filter (currently we can only hide it, not isolate it)

Pallavikumarimdb:
Introduces a new **"Own domain only"** filter on the Citations page to let users isolate search citation results belonging specifically to their own tracked domain. Activating the toggle filters both the lists and charts to show only `category = 'you'` citations, and disables the other domain-related toggles to maintain mutual exclusivity.

Closes #149

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
@@ -827,10 +827,10 @@ export default function CitationsPage() {
     if (!activeBrandId) return;
     getTopics(activeBrandId)
       .then(setTopics)
-      .catch(() => { });
+      .catch(() => {});
     getBrandPrompts(activeBrandId)
       .then((rows) => setPrompts(rows.map((r) => ({ id: r.id, text: r.text }))))
-      .catch(() => { });
+      .catch(() => {});
   }, [activeBrandId]);
 
   const loadData = useCallback(async () => {
--- a/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
@@ -156,6 +156,7 @@ interface UIFilters {
   prompt: string;
   excludeOwnDomain: boolean;
   competitorOnly: boolean;
+  ownOnly: boolean;
 }
 
 interface PromptOption {
@@ -185,6 +186,7 @@ const DEFAULT_FILTERS: UIFilters = {
   prompt: '',
   excludeOwnDomain: false,
   competitorOnly: false,
+  ownOnly: false,
 };
 
 function buildPlatformOptions(rows: CitationsOverview['rows']): PlatformOption[] {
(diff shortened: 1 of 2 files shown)


gkhngyk (approved):
Clean implementation — minimum diff, mirrors the existing two-filter pattern, and the three buttons now form a proper mutually exclusive set. Thanks @Pallavikumarimdb.

PR #172, opened 2026-06-04
fix(web): show platform insight totals

### LINKED ISSUE #169: bug(insights): per-platform Mentions/Citations show latest run instead of total
## Problem

On the Insights page (`/dashboard/insights`), expanding a prompt under **Prompt Results by Topic** reveals one row per platform. The Mentions / Citations columns in that row read the *last run's* count instead of the platform total — which makes them inconsistent with the parent rows (topic and prompt), which both show **totals**.

Repro: a prompt with one ChatGPT run at 9 mentions and a later 0-mention re-run shows:

- Topic row → Mentions: **9** (sum) ✓
- Prompt row → Mentions: **9** (sum) ✓
- Platform row → Mentions: **0** ← `latest.mentionCount`, hides the 9 ❌
- Expanding the platform → "Previous runs" table still shows the 9-mention row, so the underlying data is fine.

Result: a brand with 9 mentions in the KPI card looks like "0 mentions everywhere" at the platform level. The data is correct; only the platform-row column is reading the wrong field.

## Cause

[`web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx:989-996`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx#L989-L996):

```tsx
<p className="text-xs font-semibold tabular-nums">{latest.mentionCount}</p>
…
<p className="text-xs font-semibold tabular-nums">{latest.citationCount}</p>
```

The `PlatformGroup` interface ([line 285](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx#L285) area) already exposes `totalMentions` / `totalCitations` (sum across runs in the window), which is what the topic and prompt rows use.

## Proposed fix

Swap the two field reads in `PlatformSubGroup`:

```diff
-          <p className="text-xs font-semibold tabular-nums">{latest.mentionCount}</p>
+          <p className="text-xs font-semibold tabular-nums">{group.totalMentions}</p>
…
-          <p className="text-xs font-semibold tabular-nums">{latest.citationCount}</p>
+          <p className="text-xs font-semibold tabular-nums">{group.totalCitations}</p>
```

Leave `<VisibilityBar score={latest.visibilityScore} />` as-is — visibility is a "current state" snapshot, not a count, so showing the latest run is the right call there.

## Acceptance criteria

- [ ] Mentions / Citations on each platform row reflect the **sum** across the filtered window, matching the topic and prompt rows above.
- [ ] Per-run detail (the "Previous runs" table inside an expanded platform) is unchanged — it still shows per-run counts.
- [ ] Visibility bar still shows the latest run.
- [ ] `yarn tsc --noEmit` + `yarn lint` clean.

## Out of scope

- Reworking the platform row to add a separate "last run" badge — totals + the existing visibility bar are sufficient at this level.
- Changing the date-window semantics for "total" (the existing `groupResultsByPlatform` reduce already respects the filtered window because it operates over the results array the page already filters).

nanookclaw:
Fixes #169.

This updates the platform rows under `/dashboard/insights` so Mentions and Citations use the existing platform totals instead of the latest run counts. That keeps the platform row consistent with the topic and prompt rows above it when an older run has mentions/citations but the latest rerun is zero.

The expanded "Previous runs" table still renders each run's own counts, and the visibility bar still uses the latest run snapshot.

Verification:
- `corepack yarn install --frozen-lockfile`
- `corepack yarn --cwd web lint` (0 errors, 11 existing warnings)
- `corepack yarn --cwd web typecheck`
- `corepack yarn --cwd web format:check`
- `git diff --check`

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
@@ -988,11 +988,11 @@ function PlatformSubGroup({
         <div className="flex items-center gap-4 shrink-0">
           <div className="hidden sm:block text-right">
             <p className="text-[10px] text-muted-foreground">Mentions</p>
-            <p className="text-xs font-semibold tabular-nums">{latest.mentionCount}</p>
+            <p className="text-xs font-semibold tabular-nums">{group.totalMentions}</p>
           </div>
           <div className="hidden sm:block text-right">
             <p className="text-[10px] text-muted-foreground">Citations</p>
-            <p className="text-xs font-semibold tabular-nums">{latest.citationCount}</p>
+            <p className="text-xs font-semibold tabular-nums">{group.totalCitations}</p>
           </div>
           <SentimentBadge sentiment={latest.sentiment} />
           <div className="w-[100px] sm:w-[140px]">


gkhngyk (approved):
Minimum diff, matches the issue's proposed fix exactly. Topic / prompt / platform rows now all show totals consistently and the expanded per-run table still surfaces the latest run's count. Thanks @nanookclaw — second one this week.

Issue #187, opened 2026-06-08
fix(insights): CSV export writes platform slugs instead of display names


gkhngyk:
## Summary

On the **Answer Engine Insights** page (`/dashboard/insights`), the **Export CSV**
button produces a file whose `platform` column contains raw platform **slugs**
(e.g. `chatgpt`, `gemini`, `google-ai-mode`) instead of the human-readable
display names that the UI already shows everywhere else (e.g. `ChatGPT`,
`Gemini`, `Google AI Mode`).

## Steps to reproduce

1. Open `/dashboard/insights` for any brand with results.
2. Click **Export CSV**.
3. Open the downloaded file and look at the `platform` column.

**Expected:** display names (e.g. `ChatGPT`).
**Actual:** slugs (e.g. `chatgpt`).

## Root cause

A platform-label map already exists, but it lives locally inside the charts
component and is not shared, so the CSV export can't reach it:

- The map: `PLATFORM_LABELS` in
  [`web/src/app/[locale]/(dashboard)/dashboard/insights/_charts.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/_charts.tsx)
  — the charts render `PLATFORM_LABELS[platform] ?? platform`.
- The export writes the raw value: in `handleExportCsv` inside
  [`web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/page.tsx)
  the row maps `platform: r.platform` directly.

## Suggested fix

Move `PLATFORM_LABELS` to a shared module (e.g. under `web/src/config/`) and
use it in both places, so the chart labels and the CSV stay in sync:

```ts
platform: PLATFORM_LABELS[r.platform] ?? r.platform,
```

Falling back to the raw slug when a label is missing keeps unknown platforms
from disappearing. Please keep the existing labels intact and just centralize
the map rather than duplicating it.

PR #189, opened 2026-06-08
fix(insights): centralize platform labels for csv export

Ishant5436:
Hey, saw issue #187. I extracted PLATFORM_LABELS into a shared config file so the CSV exporter and the charts can both use it, ensuring we get the display names instead of raw slugs in the exports. I also cleaned up the duplicate definitions across the dashboard.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/agent/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/agent/page.tsx
@@ -1,3 +1,4 @@
+import { PLATFORM_LABELS } from "@/config/platform-labels";
 'use client';
 
 import { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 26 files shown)


gkhngyk:
Thanks for taking a look at this! Unfortunately this PR can't be merged as-is — a few blocking issues:

- The new `web/src/config/platform-labels.ts` was committed empty (no `PLATFORM_LABELS` export), so all the new imports resolve to `undefined`.
- The `import` was placed *above* the `'use client'` directive in many files, which breaks those client components (the directive must be the first statement in the file).
- The import was added to ~25 files that don't actually use it.
- The actual fix from #187 (making `handleExportCsv` use the display name instead of the raw slug) isn't included.

CI (format/lint/typecheck) is failing for these reasons. I'm going to close this for now, but please feel free to open a fresh, smaller PR scoped to #187 once it builds locally (`yarn lint && yarn tsc --noEmit`) — happy to review again. 🙏

PR #201, opened 2026-06-08
ci: Add lint + CI pipeline for the server package

### LINKED ISSUE #124: ci: add lint + CI pipeline for the server package
## Problem

The `server` package has **no CI coverage at all**. `.github/workflows/ci.yml` defines only a `web` job (format + typecheck). The server also has no linter configured — `server/package.json` has no `lint` script and no ESLint dependency/config, while `web` already uses ESLint + Prettier.

As a result, server code can be merged with no automated quality gate.

## Proposed change

1. Add ESLint (flat config) + Prettier to `server`, matching the conventions already used in `web` where reasonable (ESM, Node 20).
2. Add `lint` and `format:check` scripts to `server/package.json`.
3. Add a `server` job to `.github/workflows/ci.yml` that installs deps and runs lint + format check.

Example workflow job:

```yaml
  server:
    name: server — lint & format
    runs-on: ubuntu-latest
    defaults:
      run:
        working-directory: server
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v4
        with:
          node-version: 20
          cache: npm
          cache-dependency-path: server/package-lock.json
      - run: npm ci
      - run: npm run lint
      - run: npm run format:check
```

## Acceptance criteria

- [ ] ESLint + Prettier configured for `server`
- [ ] `npm run lint` and `npm run format:check` scripts added
- [ ] Server lint/format issues fixed so the repo is clean
- [ ] New `server` CI job runs on every PR and on push to `main`

## Notes

`server` uses `npm` (not `yarn`) per the README, so the job should use `npm ci`.

Pallavikumarimdb:
Adds ESLint (flat config) + Prettier matching the web package conventions for server package, new `lint` and `format:check` scripts, and a `server` job in `.github/workflows/ci.yml` that runs on every PR and push to main. Fixes all lint/format violations in existing server code.

Closes: #124


gkhngyk:
Thanks for this — really solid work. The CI job mirrors the web job cleanly, `.prettierrc.json` matches the web package exactly, the ESLint flat config is sensible, and the source changes are pure Prettier reformatting with no logic changes (checked with a whitespace-insensitive diff across the larger files). Two things to fix before merge, and one note:

**1. Don't format `src/public/t.js`.** This file is the browser tracking pixel — it's a deliberately minified single-line IIFE that gets served to every tracked page, so Prettier expanding it to ~75 lines bloats the payload and undoes the intentional minification. You already (correctly) exclude `src/public/**` from ESLint via `globalIgnores`, so let's apply the same exclusion to Prettier. Please add a `server/.prettierignore` (mirroring how the web package ignores generated files) with:
```
node_modules/
src/public/
```
and restore `src/public/t.js` to its original minified form.

**2. The `lint` script misses top-level `src/` files.** `eslint src/**/*.js` is expanded by the shell, and since `globstar` is off by default the `**` collapses to a single level — so `src/server.js` and the `src/test-*.js` files never get linted (the main server entrypoint is skipped). Since the flat config's `globalIgnores` already scopes things, simplest fix:
```json
"lint": "eslint ."
```
(`format:check` is unaffected — its glob is quoted, so Prettier does its own globstar expansion and already covers `server.js`.)

**Note on `package-lock.json`:** that change is correct and should stay — it's just the lockfile picking up the new ESLint/Prettier devDependencies, not reformatting. The new `npm ci` step in CI requires the lockfile to stay in sync with `package.json`, so reverting it would break the build.

Once t.js is excluded + restored and the lint script is fixed, this is good to merge. Thanks again!

Pallavikumarimdb:
> Thanks for this — really solid work. The CI job mirrors the web job cleanly, `.prettierrc.json` matches the web package exactly, the ESLint flat config is sensible, and the source changes are pure Prettier reformatting with no logic changes (checked with a whitespace-insensitive diff across the larger files). Two things to fix before merge, and one note:
> 
> **1. Don't format `src/public/t.js`.** This file is the browser tracking pixel — it's a deliberately minified single-line IIFE that gets served to every tracked page, so Prettier expanding it to ~75 lines bloats the payload and undoes the intentional minification. You already (correctly) exclude `src/public/**` from ESLint via `globalIgnores`, so let's apply the same exclusion to Prettier. Please add a `server/.prettierignore` (mirroring how the web package ignores generated files) with:
> 
> ```
> node_modules/
> src/public/
> ```
> 
> and restore `src/public/t.js` to its original minified form.
> 
> **2. The `lint` script misses top-level `src/` files.** `eslint src/**/*.js` is expanded by the shell, and since `globstar` is off by default the `**` collapses to a single level — so `src/server.js` and the `src/test-*.js` files never get linted (the main server entrypoint is skipped). Since the flat config's `globalIgnores` already scopes things, simplest fix:
> 
> ```json
> "lint": "eslint ."
> ```
> 
> (`format:check` is unaffected — its glob is quoted, so Prettier does its own globstar expansion and already covers `server.js`.)
> 
> **Note on `package-lock.json`:** that change is correct and should stay — it's just the lockfile picking up the new ESLint/Prettier devDependencies, not reformatting. The new `npm ci` step in CI requires the lockfile to stay in sync with `package.json`, so reverting it would break the build.
> 
> Once t.js is excluded + restored and the lint script is fixed, this is good to merge. Thanks again!

@gkhngyk I have made the requested changes, please review again.

gkhngyk:
Merged — both points addressed perfectly: `.prettierignore` excludes `src/public/` and t.js is back to its original minified form, and `lint` now uses `eslint .` so `server.js` is covered too. Server lint + CI is a great addition. Thanks for seeing it through! 🙌

PR #203, opened 2026-06-09
fix csv export platform names

Ishant5436:
Hey! the csv export was exporting raw slugs instead of the proper display names for platforms. I extracted the platform mappings into a central config file and updated the export logic across the dashboard pages to use it. Tests and lint should pass now. let me know if you need anything else!

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
@@ -30,6 +30,7 @@ import {
 } from '@/lib/actions/tracking';
 import { getTopics } from '@/lib/actions/topic';
 import type { Topic } from '@/types';
+import { getPlatformDisplayName } from '@/config/platform-labels';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
@@ -1646,7 +1647,7 @@ export default function InsightsPage() {
       created_at: r.createdAt,
       prompt: r.promptText,
       topic: r.topicName ?? '',
-      platform: r.platform,
+      platform: getPlatformDisplayName(r.platform),
       model: r.modelUsed ?? '',
       region: r.region ?? '',
       mention_count: r.mentionCount,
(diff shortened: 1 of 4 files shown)


gkhngyk:
Thanks for picking this up — the direction is exactly right (CSV should show display names, not slugs), the diff is clean, and you correctly avoided the `'use client'` pitfall. There's one correctness gap and one structural suggestion before this is ready, though.

### 1. The new map misses most real platform slugs, so the bug stays for several platforms

The actual `platform` values stored in the data (see the union in [`web/src/types/index.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/types/index.ts) and the maps used across the dashboard) include `grok`, `meta-ai`, `google-ai-overviews`, `google-ai-mode`, `google-aio`, `google-aimode`, `copilot-web`, `grok-web`, `perplexity-web`, `gemini-web`, `chatgpt-web`, etc.

`web/src/config/platform-labels.ts` only covers a handful of those and uses `google-ai` as a key, which isn't a real slug. For any unmatched slug, `getPlatformDisplayName` falls through to `|| slug`, so the CSV still exports the raw slug:

| Platform value | Result after this PR |
| --- | --- |
| `chatgpt`, `chatgpt-web`, `perplexity`, `gemini`, `claude`, `copilot` | ✅ display name |
| `grok`, `meta-ai`, `google-ai-overviews`, `google-ai-mode`, `google-aio`, `copilot-web`, `grok-web`, `perplexity-web`, `gemini-web` | ❌ still raw slug |

So Grok, Meta AI, and every Google variant would still export as slugs. CI is green only because nothing exercises the slug→label mapping yet.

### 2. A complete, correct map already exists in the file you're editing

`insights/page.tsx` already has a full `PLATFORM_LABELS` map (around lines 214–230) with all the correct slugs (`grok`, `meta-ai`, `google-ai-overviews`, `google-ai-mode`, …). The one-line fix for the export is just to reuse it:

```ts
platform: PLATFORM_LABELS[r.platform] ?? r.platform,
```

### Suggested path

This same `PLATFORM_LABELS` map is currently duplicated across ~6 dashboard files (`insights`, `insights/[id]`, `prompts/[id]`, `citations`, `competitors`, plus `_charts.tsx`). The most valuable version of this PR would be:

1. Seed `web/src/config/platform-labels.ts` from the existing **complete** map (the one in `insights/page.tsx`), so no slug is missing.
2. Replace those duplicated local maps with imports from the central file — that's the real consolidation win.
3. Then the CSV export just uses the single source.

If you'd rather keep the scope small, that's fine too — but at minimum the central map needs every real slug, otherwise the bug remains for the platforms above. Happy to re-review once it's updated. Thanks again for the contribution!

Ishant5436:
closing

Ishant5436:
hey @gkhngyk, i moved all the `PLATFORM_LABELS` maps into the central `config/platform-labels.ts` file! this also fixes the csv export since it now imports from there too. tests are passing locally, let me know if there's anything else needed.

PR #210, opened 2026-06-09
feat(auth): add password visibility toggle to auth forms

MaitreyeeDeshmukh:
## Summary

Adds a reusable `PasswordInput` component and wires it into all four auth forms so users can reveal or mask their password by clicking the eye icon.

Closes #209

## Changes

- **new** `web/src/components/ui/password-input.tsx` — reusable component that wraps the existing `Input` with an `Eye`/`EyeOff` toggle from `lucide-react` (already a dependency, no new packages added)
- `web/src/components/auth/sign-in-form.tsx` — password field swapped to `PasswordInput`
- `web/src/components/auth/sign-up-form.tsx` — password field swapped to `PasswordInput`
- `web/src/components/auth/reset-password-form.tsx` — both password fields swapped to `PasswordInput`, `Input` import removed (no longer needed)
- `web/src/components/team/accept-invitation-card.tsx` — both password fields swapped to `PasswordInput`

## Acceptance criteria

- [x] Password fields have a working eye toggle (masked by default)
- [x] Toggle button is `type="button"` — does not submit the form
- [x] `aria-label` set for accessibility (`Show password` / `Hide password`)
- [x] No existing form logic, validation, or state changed
- [x] TypeScript: zero errors
- [x] Lint: zero new warnings

PR #212, opened 2026-06-09
fix(ui): merge caller className in PasswordInput via cn

### LINKED ISSUE #211: fix(ui): PasswordInput drops caller-provided className
## What

`PasswordInput` (added in #210) hardcodes `className="pr-10"` **after** spreading `{...props}`, so a `className` passed by a caller is silently dropped instead of being merged.

[`web/src/components/ui/password-input.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/ui/password-input.tsx):

```tsx
<Input type={visible ? 'text' : 'password'} {...props} className="pr-10" />
```

Because `className="pr-10"` comes last, it overrides `props.className`. None of the current four callers pass a `className`, so there's **no visible bug today** — but it's a reusable `ui/` component, so the first caller that needs e.g. an error / `aria-invalid` style will lose it.

## Fix

Destructure `className` out of props and merge it with the base classes via `cn`:

```tsx
import { cn } from '@/lib/utils';

export function PasswordInput({ className, ...props }: PasswordInputProps) {
  const [visible, setVisible] = useState(false);
  return (
    <div className="relative">
      <Input
        type={visible ? 'text' : 'password'}
        className={cn('pr-10', className)}
        {...props}
      />
      {/* ...toggle button unchanged... */}
    </div>
  );
}
```

## Acceptance

- [ ] A `className` passed to `PasswordInput` is merged with `pr-10` (both applied), not dropped
- [ ] The four existing auth forms render unchanged
- [ ] No new TypeScript / lint errors

## Notes

- Fix-forward from #210 (the toggle feature itself is already merged and working).
- UI-only, low-risk change.

MaitreyeeDeshmukh:
## Summary

Fixes a bug in the `PasswordInput` component where a `className` passed by a caller was silently dropped instead of being merged.

Fixes #211
Follow-up to #210

## Root cause

`className="pr-10"` was placed after `{...props}` on the `Input` element, so any `className` in props was overridden. No current caller passes a `className` so there is no visible bug today, but as a reusable `ui/` component the first caller that needs an error style or `aria-invalid` class would silently lose it.

## Change

- Destructure `className` out of props in the function signature
- Merge with `cn('pr-10', className)` so `pr-10` is always applied as the base and the caller's class is appended
- Place `className` before `{...props}` so the spread cannot override it

## Acceptance criteria

- [x] A `className` passed to `PasswordInput` is merged with `pr-10` (both applied), not dropped
- [x] The four existing auth forms render unchanged (none pass a `className`)
- [x] TypeScript: zero errors
- [x] Lint: zero new errors

### DIFF
--- a/web/src/components/ui/password-input.tsx
+++ b/web/src/components/ui/password-input.tsx
@@ -3,15 +3,16 @@
 import { useState } from 'react';
 import { Eye, EyeOff } from 'lucide-react';
 import { Input } from '@/components/ui/input';
+import { cn } from '@/lib/utils';
 
 type PasswordInputProps = Omit<React.ComponentProps<typeof Input>, 'type'>;
 
-export function PasswordInput(props: PasswordInputProps) {
+export function PasswordInput({ className, ...props }: PasswordInputProps) {
   const [visible, setVisible] = useState(false);
 
   return (
     <div className="relative">
-      <Input type={visible ? 'text' : 'password'} {...props} className="pr-10" />
+      <Input type={visible ? 'text' : 'password'} className={cn('pr-10', className)} {...props} />
       <button
         type="button"
         onClick={() => setVisible((v) => !v)}


gkhngyk:
Merged — exactly the fix: `cn('pr-10', className)` with `className` pulled out of props so callers' classes are preserved. Clean follow-up, thanks again! 🙌

PR #235, opened 2026-06-11
fix(web): group insights results by platform

VrtxOmega:
## Summary

- Group the insights dashboard prompt-result buckets by platform only, so provider model slug changes do not split one logical platform into duplicate rows.
- Keep the platform group's model metadata sourced from the latest run, so the badge still reflects the current model.
- Move the pure grouping helpers into a small local module and add Vitest coverage for slug drift plus distinct-platform separation.

## Related issue

Closes #231

## Type of change

- [ ] `feat` - New feature
- [x] `fix` - Bug fix
- [ ] `chore` - Maintenance / dependencies
- [ ] `docs` - Documentation only
- [ ] `refactor` - Code change that neither fixes a bug nor adds a feature
- [x] `test` - Adding or updating tests

## Validation

- `corepack yarn test 'src/app/[locale]/(dashboard)/dashboard/insights/grouping.test.ts'`
- `corepack yarn prettier --check 'src/app/[locale]/(dashboard)/dashboard/insights/page.tsx' 'src/app/[locale]/(dashboard)/dashboard/insights/grouping.ts' 'src/app/[locale]/(dashboard)/dashboard/insights/grouping.test.ts'`
- `corepack yarn eslint 'src/app/[locale]/(dashboard)/dashboard/insights/page.tsx' 'src/app/[locale]/(dashboard)/dashboard/insights/grouping.ts' 'src/app/[locale]/(dashboard)/dashboard/insights/grouping.test.ts'`
- `corepack yarn format:check`
- `corepack yarn test`
- `corepack yarn typecheck`
- `corepack yarn lint` (0 errors; existing unrelated warnings remain in other files)
- `git diff --cached --check`

## Checklist

- [x] Branch follows the naming convention (`feature/`, `fix/`, `chore/`, `docs/`) - see [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md)
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] `yarn lint` passes (run from `web/`)
- [x] `yarn typecheck` passes (run from `web/`)
- [x] `yarn format:check` passes (run from `web/`)
- [x] Changes are focused - one concern per PR

PR #237, opened 2026-06-11
fix(web): group prompt detail results by platform

### LINKED ISSUE #236: bug(prompts): duplicate platform groups on the prompt detail page when a model slug changes
## Problem

On the prompt detail page (`/dashboard/prompts/<id>`), the per-platform grouping splits one logical platform into **two groups** when the provider's model slug changes over time — e.g. ChatGPT showing up twice because its model went from `gpt-5-5` to `gpt-5-3-mini`.

This is the same bug that was just fixed for the **Answer Engine Insights** page in #231 / #235 — the prompt detail page has its own copy of the grouping logic that was missed.

## Steps to reproduce

1. Open a prompt detail page for a prompt that has been tracked across a model-slug change (e.g. ChatGPT or Grok, whose slugs have drifted).
2. Observe the platform groups: the same platform appears as two separate cards.

**Expected:** one group per platform, with the model badge reflecting the latest run.
**Actual:** two groups for the same platform, one per historical model slug.

## Root cause

`groupByPlatform()` in [`web/src/app/[locale]/(dashboard)/dashboard/prompts/[id]/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/%5Bid%5D/page.tsx) keys groups by **platform + model**:

```ts
const key = `${result.platform}|${result.modelUsed ?? ''}`;
```

`modelUsed` carries whatever slug the provider returned, so each slug change spawns a phantom second group. A platform is never tracked with two models concurrently — the slug just drifts over time.

## Suggested fix

Mirror the fix from #235: key the groups by **platform only**, and keep the group's `modelUsed` sourced from the latest run so the badge still shows the current model.

```ts
const key = result.platform;
```

The function already sets `modelUsed: latest.modelUsed` from the most-recent run (line ~154), so the badge keeps reflecting the current model with no further change. Genuinely distinct platforms keep distinct `platform` values, so they stay separate.

## Acceptance criteria

- [ ] A prompt whose platform model slug changed over time shows **one** group per platform (no duplicate ChatGPT / Grok cards)
- [ ] The platform group's model badge reflects the **most recent** run's model
- [ ] The previous-runs history inside a group spans runs from before and after the slug change
- [ ] Genuinely distinct platforms remain separate groups
- [ ] Display/grouping layer only — no data or RPC changes

## Notes

The insights page now keeps this logic in a shared, unit-tested `grouping.ts` module (#235). This prompt-detail copy could either reuse that pattern or apply the same one-line key change locally — whichever the contributor prefers; a small Vitest case for the slug-drift scenario would be welcome.

VrtxOmega:
## Summary
- Extract the prompt detail page platform grouping helper into a local module, mirroring the accepted insights grouping pattern from #235.
- Group prompt detail results by `platform` only, so provider model slug changes do not split one logical platform into duplicate cards.
- Keep the group model/region metadata sourced from the latest run and add Vitest coverage for slug drift plus distinct-platform separation.

## Related issue

Closes #236

## Why

The prompt detail page had its own `groupByPlatform()` copy that keyed groups by `platform|modelUsed`. When a provider slug drifted over time, the same platform could appear as two cards even though it was the same logical platform history.

## Validation

- `corepack yarn install --frozen-lockfile`
- `corepack yarn test 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.test.ts'`
- `corepack yarn prettier --check 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/page.tsx' 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.ts' 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.test.ts'`
- `corepack yarn eslint 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/page.tsx' 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.ts' 'src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.test.ts'`
- `corepack yarn test`
- `corepack yarn typecheck`
- `corepack yarn format:check`
- `corepack yarn lint` (0 errors; existing unrelated warnings remain in other files)
- `git diff --cached --check`
- `git diff --check HEAD~1..HEAD`

## Scope

Display/grouping layer only. No data fetching, RPC, or stored data changes.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.test.ts
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/[id]/grouping.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it } from 'vitest';
+import type { PromptResultWithText } from '@/lib/actions/tracking';
+import { groupByPlatform } from './grouping';
+
+function result(overrides: Partial<PromptResultWithText>): PromptResultWithText {
+  return {
+    id: 'result-id',
+    promptId: 'prompt-id',
+    brandId: 'brand-id',
+    platform: 'chatgpt-web',
+    response: 'response text',
+    citations: [],
+    mentionCount: 0,
+    citationCount: 0,
+    sentiment: 'neutral',
+    visibilityScore: 0,
+    createdAt: '2026-06-01T00:00:00.000Z',
+    promptText: 'Where should I buy running shoes?',
+    ...overrides,
+  };
+}
+
+describe('groupByPlatform', () => {
+  it('keeps model slug drift in one platform group and uses latest metadata', () => {
+    const groups = groupByPlatform([
+      result({
+        id: 'older-run',
+        modelUsed: 'gpt-5-5',
+        region: 'US',
+        mentionCount: 1,
+        citationCount: 2,
+        visibilityScore: 40,
+        createdAt: '2026-06-01T00:00:00.000Z',
+      }),
+      result({
+        id: 'latest-run',
+        modelUsed: 'gpt-5-3-mini',
+        region: 'CA',
+        mentionCount: 3,
+        citationCount: 4,
+        visibilityScore: 80,
+        createdAt: '2026-06-02T00:00:00.000Z',
+      }),
+    ]);
+
+    expect(groups).toHaveLength(1);
+    expect(groups[0]).toMatchObject({
+      key: 'chatgpt-web',
+      platform: 'chatgpt-web',
+      modelUsed: 'gpt-5-3-mini',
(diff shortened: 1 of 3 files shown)


VrtxOmega:
Thanks for the quick review and merge. I appreciate you pointing me to the sibling prompt-detail case after #235 — it made sense to keep this follow-up narrow and mirror the tested grouping approach there.

Happy to help again if another similar UI grouping edge case shows up.

PR #267, opened 2026-06-16
fix(audit): hub recent-audits list survives a failing quota/trend call

gkhngyk:
## Problem

The Site Audit hub loaded recent audits, the monthly quota, and the trend in one `Promise.all`. If quota or trend failed — which happens during a deploy window where the **web is live but the server hasn't been redeployed** with the new `/api/audits/quota` & `/api/audits/trend` routes — the whole batch rejected and the page showed **neither the recent audits nor the chart** (data wasn't lost, just not rendered).

## Fix

Fetch each independently with its own try/catch: recent audits load on their own; quota and trend are best-effort enhancements that hide quietly if their (newer) endpoints aren't available yet. No more all-or-nothing.

Web-only.


gkhngyk:
Closing — the live issue was the server not being redeployed (now fixed). This was just optional hardening for the web-deploys-before-server window; not worth the noise right now.

PR #277, opened 2026-06-19
docs(env): add CRON_SECRET to both .env.example files

### LINKED ISSUE #244: docs(env): CRON_SECRET missing from .env.example (internal endpoints 401 in cloud)
## Summary

`CRON_SECRET` is required to operate the app in cloud mode, but it is **missing from both `.env.example` files** — so nothing reminds an operator to set it, and the internal endpoints permanently return `401`.

### Where it's used

**Server — guards three internal endpoints** ([`server/src/server.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/server.js)):

- `server.js:155` — `/api/internal/daily-tracking`
- `server.js:177` — `/api/internal/content/:id/brief`
- `server.js:213` — `/api/internal/trigger-tracking`

Each guard is `if (!process.env.CRON_SECRET || secret !== process.env.CRON_SECRET) → 401`. When the var is unset, the **first clause** is always true, so every call to these endpoints returns `401` no matter what the caller sends.

These are documented as `CRON_SECRET`-protected in [`server/README.md`](https://github.com/ansvisor/ansvisor/blob/main/server/README.md) (the internal-endpoints table).

**Web — sends / checks the same secret** (must match the server's value):

- `web/src/app/api/cron/daily-tracking/route.ts` — Vercel Cron entrypoint
- `web/src/app/api/stripe/success/route.ts` — post-checkout tracking trigger
- `web/src/lib/mcp/data.ts` — MCP brief generation (throws `CRON_SECRET must be configured` when missing)

### Impact

In a cloud deploy, if the operator never sets `CRON_SECRET` (because it isn't in the example files), daily tracking cron, content-brief generation, and post-checkout tracking all fail — the server side 401s and the MCP path throws. Self-hosted/local users who never hit these paths won't notice, which is why it's easy to miss.

## Suggested fix

Add a documented `CRON_SECRET` entry to **both** example files:

- [`server/.env.example`](https://github.com/ansvisor/ansvisor/blob/main/server/.env.example)
- [`web/.env.example`](https://github.com/ansvisor/ansvisor/blob/main/web/.env.example)

Include a comment noting that:

- it's **cloud-only** (needed when Vercel Cron / the Stripe success route / the MCP layer call the server's internal endpoints), and
- the **same value must be set on both the web app and the server**, since the web side sends `Authorization: Bearer ${CRON_SECRET}` and the server compares against its own copy.

Docs/config only — no code changes.

P-Maheswari:
Fixes #244. CRON_SECRET guards three internal server endpoints but was missing from both example files, causing silent 401s in cloud deployments. Added with a comment explaining it is cloud-only and must be set to the same value on both the web app and the server.

## Summary

Added `CRON_SECRET` to both `.env.example` files with a descriptive comment.
Without this entry, cloud operators have no indication they need to set this
variable, causing permanent 401s on three internal endpoints and failures in
daily tracking, content-brief generation, and post-checkout tracking.

## Related issue

Closes #244

## Type of change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix
- [x] `docs` — Documentation only

## Checklist

> `lint`, `typecheck`, `format:check` skipped — docs/config only change, no source files modified.

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -62,6 +62,17 @@ PLATFORM_PROVIDER=cloro
 # AI Volume Multiplier (AI search adoption rate, e.g. 0.15 = 15%)
 AI_VOLUME_MULTIPLIER=0.15
 
+# Cron / Internal-endpoint secret (cloud only — not needed for self-hosted)
+# Guards three internal server endpoints:
+#   /api/internal/daily-tracking
+#   /api/internal/content/:id/brief
+#   /api/internal/trigger-tracking
+# The web app sends this as "Authorization: Bearer <value>"; the server
+# compares against its own copy. MUST be the same value on both sides.
+# Generate: openssl rand -hex 32
+# CRON_SECRET=
+
 # Stripe (cloud only — not needed for self-hosted)
 # STRIPE_SECRET_KEY=sk_test_...
 # STRIPE_WEBHOOK_SECRET=whsec_...
+
--- a/web/.env.example
+++ b/web/.env.example
@@ -59,6 +59,19 @@ NEXT_PUBLIC_IS_CLOUD=false
 NEXT_PUBLIC_API_URL=http://localhost:80
 
 # -----------------------------------------------------------------------------
+# Cron / Internal-endpoint secret (cloud only — not needed for self-hosted)
+# -----------------------------------------------------------------------------
+# Shared secret that authenticates calls from the web layer to the server's
+# internal endpoints. Used by:
+#   - Vercel Cron  →  /api/cron/daily-tracking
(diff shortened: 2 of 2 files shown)


gkhngyk (approved):
Verified against the code — all three internal endpoints (`/api/internal/daily-tracking`, `/api/internal/content/:id/brief`, `/api/internal/trigger-tracking`) are guarded by `CRON_SECRET`, and it was indeed missing from both example files. The comments are accurate and the cloud-only framing is spot on. Thanks @P-Maheswari! 🙏

PR #302, opened 2026-06-26
refactor(server): consolidate intent-extraction logic into a single module

### LINKED ISSUE #241: refactor(server): consolidate intent-extraction logic into a single module
## Summary

The intent-extraction logic is copy-pasted between the volumes route and the manual smoke script, and the two copies have **already drifted apart** — so the script no longer tests what production actually does.

Three definitions are duplicated:

- `intentKeywordSchema`
- `INTENT_SYSTEM_PROMPT`
- `AI_VOLUME_MULTIPLIER`

in both [`server/src/routes/volumes.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/volumes.js) and [`server/src/test-volumes.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/test-volumes.js).

### The drift

| | `routes/volumes.js` (production) | `test-volumes.js` (smoke) |
|---|---|---|
| Prompt | "5 short **HEAD** keywords, 1–3 words (strict)", with worked examples and a "strip modifiers" rule | "5 keyword phrases, **2–5 words each**", no examples |
| Schema `intent` enum | 9 values (includes `other`) | 7 values |
| keyword field | described as a Google head term with measurable volume | undescribed |

Production deliberately asks for short head terms (because Google Ads returns 0 volume for long-tail strings); the smoke script still asks for 2–5 word phrases. The script tests a prompt/schema that production no longer uses.

### Duplication inside the route, too

The same `generateObject({ model, schema: intentKeywordSchema, system: INTENT_SYSTEM_PROMPT, prompt: ... })` call appears twice in `volumes.js` — at `volumes.js:201` and `volumes.js:321` — with identical arguments.

## Suggested fix

Extract a single source of truth, e.g. `server/src/lib/intent-extraction.js`, exporting:

- `intentKeywordSchema`
- `INTENT_SYSTEM_PROMPT`
- `AI_VOLUME_MULTIPLIER`
- a small helper, e.g. `extractIntentKeywords(promptText, model)`, that wraps the shared `generateObject` call and returns `{ intent, keywords }`.

Then:

- `routes/volumes.js` imports from the module and both call sites (lines 201, 321) collapse to one `await extractIntentKeywords(...)`.
- If the smoke script survives (see #240), it imports the same module instead of carrying its own copy — drift becomes impossible.

## Relationship to #240

This overlaps with #240 (move/remove the manual smoke scripts):

- If `test-volumes.js` is **deleted** there, the cross-file duplication disappears and this issue narrows to the in-route dedup (lines 201/321) plus the lib extraction.
- If it is **moved and kept**, this extraction is what lets it import the canonical definitions instead of a drifted copy.

Either way, the production-side extraction stands on its own and is worth doing.

## Type

`refactor` — no behavior change to the volumes endpoint; production keeps its current (head-term) prompt and schema.

gkhngyk:
## Summary

Consolidates the duplicated intent-extraction logic in the volumes routes into a single module.

The schema, system prompt, multiplier, and the `generateObject` call used to extract search intent + 5 Google head keywords were defined inline in `server/src/routes/volumes.js`, and the same `generateObject({ schema, system, prompt })` call appeared twice — in `/analyze` and `/analyze-batch` — with identical arguments.

This PR extracts a single source of truth, `server/src/lib/intent-extraction.js`, exporting:

- `intentKeywordSchema`
- `INTENT_SYSTEM_PROMPT`
- `AI_VOLUME_MULTIPLIER`
- `extractIntentKeywords(promptText, aiModel)` — wraps the shared `generateObject` call and returns `{ intent, keywords }`.

Both call sites in `volumes.js` now collapse to a single `await extractIntentKeywords(...)`, and the now-unused `ai` / `zod` imports are dropped from the route.

> Note: the issue also mentioned cross-file drift with `server/src/test-volumes.js`. That smoke script was already deleted in #256, so the cross-file duplication no longer exists — this PR covers the in-route dedup + the lib extraction, which stands on its own.

No behavior change: production keeps its current head-term prompt and schema.

## Related issue

Closes #241

## Type of change

## Checklist

- [x] Branch follows the naming convention (`refactor/`)
- [x] Server lint passes (`eslint`) on changed files
- [x] Server `prettier --check` passes on changed files
- [x] Server test suite passes (`vitest run` — 93/93)

### DIFF
--- a/server/src/lib/intent-extraction.js
+++ b/server/src/lib/intent-extraction.js
@@ -1,7 +1,8 @@
 import { generateObject } from 'ai';
 import { z } from 'zod';
 
-export const AI_VOLUME_MULTIPLIER = parseFloat(process.env.AI_VOLUME_MULTIPLIER || '0.15');
+const parsedMultiplier = parseFloat(process.env.AI_VOLUME_MULTIPLIER);
+export const AI_VOLUME_MULTIPLIER = Number.isFinite(parsedMultiplier) ? parsedMultiplier : 0.15;
 
 export const intentKeywordSchema = z.object({
   intent: z
--- a/server/src/lib/intent-extraction.js
+++ b/server/src/lib/intent-extraction.js
@@ -1,0 +1,74 @@
+import { generateObject } from 'ai';
+import { z } from 'zod';
+
+export const AI_VOLUME_MULTIPLIER = parseFloat(process.env.AI_VOLUME_MULTIPLIER || '0.15');
+
+export const intentKeywordSchema = z.object({
+  intent: z
+    .enum([
+      'comparison',
+      'how-to',
+      'what-is',
+      'best-top',
+      'vs-review',
+      'recommendation',
+      'problem-solving',
+      'other',
+    ])
+    .describe('The primary search intent behind this prompt'),
+  keywords: z
+    .array(
+      z
+        .string()
+        .describe(
+          'A short Google head term — 1 to 3 words, category-level, measurable Google Ads volume',
+        ),
+    )
+    .length(5)
+    .describe(
+      'Five broad, high-volume Google head terms that capture the category the prompt belongs to',
+    ),
+});
+
+export const INTENT_SYSTEM_PROMPT = `You are a keyword research expert. Given an AI search prompt, you must:
+
+1. Determine the primary search intent (comparison, how-to, what-is, best-top, vs-review, recommendation, or problem-solving).
+2. Generate exactly 5 short HEAD keywords that a user would type into Google when researching the same category.
+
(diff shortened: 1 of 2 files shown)


gkhngyk:
Thanks @Copilot — addressed:

**1. `AI_VOLUME_MULTIPLIER` NaN guard ✅ Fixed** (c2f813c). Now parses then falls back to `0.15` when the value isn't finite, so a non-numeric/empty env var can't poison `estAiVolume` or persisted rows. Verified: `abc` → `0.15`, `0.2` → `0.2`, unset → `0.15`.

**2 & 3. System-prompt wording (missing `other`; "1–3 words strict" vs "never more than 4") — intentionally not changed in this PR.** This prompt was moved **verbatim** from the existing production code; both quirks are pre-existing, not introduced here. This PR is a pure `refactor` whose stated contract is *no behavior change to the volumes endpoints — production keeps its current prompt and schema*. Editing the prompt text would change the LLM's output, which is out of scope. They're fair observations and can be picked up as a separate prompt-tuning change if the team wants to refine extraction behavior.

PR #315, opened 2026-06-28
Show readable region names in the insights filter

### LINKED ISSUE #314: feat(insights): show flags + country names in the Region filter instead of raw codes
## What

The Region filter on [`/dashboard/insights`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx) shows raw region codes (`US`, `GB`, `DE`, …). Show a **flag + readable country name** next to each option (e.g. `🇺🇸 United States`) so the picker is easier to scan.

## Current state

The select renders the bare code:

```tsx
{availableRegions.map((r) => (
  <SelectItem key={r} value={r}>
    {r}
  </SelectItem>
))}
```

`availableRegions` are ISO 3166-1 alpha-2 codes derived from `prompt_results.region`. The repo already ships a code → full-name map in [`web/src/config/prompt-options.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/config/prompt-options.ts) (`REGIONS`: `{ code: 'US', label: 'United States' }`, …).

## Proposed change

1. **Add a tiny region-display helper** (e.g. `web/src/lib/region.ts`):
   - `regionFlag(code)` — derive the flag emoji from the ISO code with no library or asset: map each of the two letters to its regional-indicator symbol (`String.fromCodePoint(0x1f1e6 + (letter.charCodeAt(0) - 65))`).
   - `regionLabel(code)` — full name from `REGIONS`, falling back to the raw code if it isn't in the list.
2. **Use it in the insights Region select** — show `🇺🇸 United States` (or `🇺🇸 US`) in both the dropdown items and the trigger's selected value. Keep `All Regions` as-is.
3. **Reuse the helper anywhere else regions render** for consistency (e.g. any region picker in onboarding / brand settings / prompt add), but the insights filter is the primary target.

## Notes / edge cases

- Codes not present in `REGIONS` or that aren't valid two-letter codes should fall back gracefully (show the code, skip the flag) — don't crash on unexpected values.
- Flag emoji rendering depends on the OS font: most platforms render them, but Windows shows the two letters instead of a flag. Acceptable — the readable country name still makes the option clear. Worth a one-line note in the PR.
- UI-only change; no data, schema, or server changes.

## How to test

1. On `/dashboard/insights`, open the Region filter on a brand that has results in more than one region → each option shows a flag + country name; selecting one shows the same in the trigger.
2. A region code that isn't in `REGIONS` still renders (code only, no crash).
3. `All Regions` and the existing filter behavior are unchanged.

Peter7896:
Closes #314.

This adds a small region display helper backed by the existing `REGIONS` config and uses it in the insights Region filter. Known ISO region codes now render as flag plus country name in both the selected value and menu items, while unknown or invalid values fall back to the raw code.

Validation:
- `node_modules/.bin/vitest.cmd run src/lib/region.test.ts` -> 3 passed
- `node_modules/.bin/tsc.cmd --noEmit`
- `node_modules/.bin/prettier.cmd --check src/**/*.{ts,tsx}`
- `git diff --check`

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
@@ -56,6 +56,7 @@ import {
 } from '@/lib/actions/tracking';
 import { getTopics } from '@/lib/actions/topic';
 import { PLATFORM_LABELS } from '@/config/platform-labels';
+import { formatRegionDisplay } from '@/lib/region';
 import type { Topic } from '@/types';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Badge } from '@/components/ui/badge';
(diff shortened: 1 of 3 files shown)


gkhngyk:
Merged — thanks for the clean contribution, @Peter7896! 🎉 Nice touch reusing the existing `REGIONS` config and deriving the flags from the ISO codes with no extra dependency, plus the fallback handling and tests. Looking forward to more from you!

PR #320, opened 2026-06-29
Sort metric breakdown rows by current value

### LINKED ISSUE #316: feat(insights): sort the metric breakdown drawer's prompts list by current value (Cur) descending
## What

In [`/dashboard/insights`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx), clicking a metric card (e.g. Mentions) opens the **metric breakdown drawer** ([`_metric-breakdown-sheet.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/_metric-breakdown-sheet.tsx)). On the **Prompts** tab, the list should be ordered by the **Cur** (current value) column **descending** — highest current value at the top.

## Current behavior

The table renders the rows in the order it receives them, and that order is **ascending by `delta`** (biggest drop first), not by current value. The sort is applied once in the data builder, [`getMetricBreakdown` in `web/src/lib/actions/tracking.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts):

```ts
rows.sort((a, b) => a.delta - b.delta); // ascending by delta — byPrompt[0] = largest drop
```

So a prompt with a high current mention count can appear far down the list if its delta is small/positive.

## Why the fix belongs in the view, not the data builder

The delta-ascending order is **intentionally relied on** by the drawer's "biggest driver" sentence (`TopContributorLine`), which reads `data.byPrompt[0]` (largest drop) and `data.byPrompt[data.byPrompt.length - 1]` (largest gain). Re-sorting the shared array in `getMetricBreakdown` would silently break that attribution. So the display order should be changed **in the sheet**, leaving the action's array order intact.

## Proposed change

In `_metric-breakdown-sheet.tsx`, sort the rows **by `cur` descending for display** — e.g. in `BreakdownTable` (or where `pickRows()` returns), map over a sorted **copy** (`[...rows].sort((a, b) => b.cur - a.cur)`) so the table shows highest current value first. Leave `getMetricBreakdown`'s ordering untouched so `TopContributorLine` still works.

- Primary target: the **Prompts** tab (the user's request).
- Since the table is shared, applying the same `cur`-descending sort to the Platforms / Topics tabs is fine and keeps the three tabs consistent.
- Optional: a stable tiebreak for equal `cur` (e.g. by `delta` desc or label) to keep ordering deterministic.

## Acceptance criteria

- [ ] In the breakdown drawer's Prompts tab, rows are ordered by the **Cur** column from highest to lowest.
- [ ] The "biggest driver" sentence at the top of the drawer is unchanged (still names the largest drop/gain correctly).
- [ ] No change to the underlying data or to `getMetricBreakdown`'s returned order.
- [ ] (If applied) Platforms and Topics tabs follow the same Cur-descending order.

## Notes

- View-only change; no data, schema, or server changes.

Peter7896:
This fixes #316 by sorting the metric breakdown drawer rows by current value for display without changing the underlying breakdown order used by the top-driver summary.

What changed:
- Added a small display helper that returns a sorted copy of breakdown rows by `cur` descending, with deterministic tie-breakers.
- Updated the breakdown table to render that display order for prompts, platforms, and topics.
- Added focused coverage proving the helper sorts by current value and does not mutate the source row order.

Validation:
- `npm test -- breakdown-display.test.ts`
- `npm run typecheck`
- `npm run format:check`
- `git diff --check`

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/insights/_metric-breakdown-sheet.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/insights/_metric-breakdown-sheet.tsx
@@ -20,6 +20,7 @@ import {
   type BreakdownRow,
   type InsightsBreakdown,
 } from '@/lib/actions/tracking';
+import { sortBreakdownRowsForDisplay } from './breakdown-display';
 
 const METRIC_TITLE: Record<BreakdownMetric, string> = {
   mentions: 'Mentions',
@@ -259,7 +260,7 @@ function BreakdownTable({
           </tr>
         </thead>
         <tbody>
-          {rows.map((r) => (
+          {sortBreakdownRowsForDisplay(rows).map((r) => (
             <tr key={r.id} className="border-b last:border-b-0 hover:bg-muted/30">
               <td className="px-3 py-2.5 min-w-0">
                 {kind === 'prompts' ? (
--- a/web/src/app/[locale]/(dashboard)/dashboard/insights/breakdown-display.test.ts
+++ b/web/src/app/[locale]/(dashboard)/dashboard/insights/breakdown-display.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from 'vitest';
+import type { BreakdownRow } from '@/lib/actions/tracking';
+import { sortBreakdownRowsForDisplay } from './breakdown-display';
+
+function row(id: string, cur: number, delta = 0, label = id): BreakdownRow {
+  return {
+    id,
+    label,
+    prev: cur - delta,
+    cur,
+    delta,
+    deltaPct: null,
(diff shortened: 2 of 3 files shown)


gkhngyk (approved):
Thanks again, @Peter7896 — another clean one! 🙌 I like that you kept it a display-only sort on a copy (`[...rows].sort`), so the underlying delta order the top-driver summary depends on stays intact, and the deterministic tie-breakers + non-mutation test are exactly the right touches. Verified locally: typecheck, lint, and the new test all green. Merging now.

PR #339, opened 2026-07-04
Rebrand/optumus cpanel

Joseymras:


gkhngyk:
Thanks for your interest in the project! This PR rebrands the entire codebase from Ansvisor to "Optumus Analytics" (name, URLs, GitHub org, docs, demo credentials, Docker images, MCP/skill names, etc.). That's a downstream/fork-level change — it's specific to running your own branded instance, so it isn't something we can merge into upstream ansvisor/ansvisor.

PR #355, opened 2026-07-06
feat(prompts): paginate High frequency table 10/page with numbered controls (#349)

### LINKED ISSUE #349: feat(prompts): paginate the Query Fan-out "High frequency" table (10 per page)
## Summary

Turn the current Query Fan-out table into a **High frequency** table that pages 10 rows at a time. Today the tab renders every observed sub-query in one long list; on brands with a lot of fan-out that's a wall of rows. It's already sorted by **Times searched** (most-searched first), so paginating at 10/page means page 1 is the 10 highest-frequency sub-queries, page 2 the next 10, and so on, with numbered page controls (1 2 3 4 …).

## What to change

- Frame the existing table on the **Query Fan-out** tab as **High frequency** (heading/label), keeping the default sort by **Times searched** descending.
- Paginate at **10 rows per page** with numbered page navigation.
- Reset to page 1 when the active brand changes and after a successful track (so the flipped/updated list starts from the top).

## Where

- [`web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/_fanout-tab.tsx) — the table + a `page` state + a page-controls footer.
- [`web/src/lib/actions/fanout.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/fanout.ts) — `getQueryFanout` already returns `subQueries` sorted by `timesSearched` desc, so pagination can be **client-side** (slice the array by page). No action change needed for v1.

## Notes

- Fan-out is sparse (mostly Copilot/Perplexity, per #332), so client-side pagination over the already-sorted array is fine; move to server-side paging only if a brand's list grows large enough to matter.
- Pairs with the sibling "Query Fanouts by Prompts" view (separate issue) — the two are the two lenses of the Query Fan-out tab (a "High frequency" / "By prompt" toggle would let them share the tab).

## How to test

1. On a brand with more than 10 observed sub-queries, `/dashboard/prompts?tab=fanout` shows 10 rows with page controls (1 2 3 4 …); page 1 is the 10 highest **Times searched**.
2. Paging forward/back shows the next/previous 10 in the same sort order.
3. Switching brand or tracking a sub-query resets to page 1.

Maqbool61:
Closes #349

## What changed
All changes are contained in a single file: `_fanout-tab.tsx`.

### 1. Renamed heading
- `CardTitle` text changed from `"Query fan-out"` → `"High frequency"` as specified in the issue.

### 2. Client-side pagination (10/page)
- Added `page` state (`useState(1)`) and a `PAGE_SIZE = 10` constant.
- `data.subQueries` is sliced into `pageRows` using `(page - 1) * PAGE_SIZE` offset before being passed to the table — no server changes needed.

### 3. Numbered page controls
- Rendered below the table when `totalPages > 1`.
- Each page number is a `Button` from the existing shadcn/ui component.
- Active page uses `variant="default"`, others use `variant="ghost"` for clear visual distinction.

### 4. Page reset on brand change
- Added a `useEffect` watching `brandId` that calls `setPage(1)` — ensures switching brands always starts at page 1.

### 5. Page reset after successful track
- `setPage(1)` called inside `handleTrack` immediately after `toast.success` — ensures the user lands back on page 1 after tracking a query.

## What was NOT changed
- No server actions touched (`fanout.ts` unchanged).
- No other files modified.
- Zero new TypeScript errors introduced (verified via `tsc --noEmit` diff before/after).

## Testing
- Load the prompts page with a brand that has >10 fan-out sub-queries → numbered controls appear below the table.
- Click a page number → table updates to show the correct slice.
- Switch brand → page resets to 1.
- Track a query → page resets to 1 after success toast.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
@@ -120,72 +120,74 @@ export function QueryFanoutTab({ brandId }: { brandId: string }) {
                     <TableHeader>
                       <TableRow>
                         <TableHead>Sub-query</TableHead>
-                <TableHead className="w-[160px]">Engine</TableHead>
-                <TableHead className="w-[120px] text-right">Times searched</TableHead>
-                <TableHead className="w-[220px]">Sourced prompts</TableHead>
-                <TableHead className="w-[130px]">Intent</TableHead>
-                <TableHead className="w-[64px] text-right">Track</TableHead>
-              </TableRow>
-            </TableHeader>
-            <TableBody>
-              {pageRows.map((sq) => {
-                const key = sq.query.toLowerCase();
-                const adding = addingKey === key;
-                return (
-                  <TableRow key={key}>
-                    <TableCell className="font-medium">{sq.query}</TableCell>
-                    <TableCell>
-                      <div className="flex flex-wrap gap-1">
-                        {sq.engines.map((e) => (
-                          <Badge key={e} variant="secondary" className="text-[10px]">
-                            {platformLabel(e)}
-                          </Badge>
-                        ))}
-                      </div>
-                    </TableCell>
-                    <TableCell className="text-right tabular-nums">{sq.timesSearched}</TableCell>
-                    <TableCell>
-                      <SourcedPrompts prompts={sq.sourcedPrompts} />
-                    </TableCell>
-                    <TableCell>
-                      <IntentBadge intent={intents[key]} />
-                    </TableCell>
-                    <TableCell className="text-right">
-                      {sq.tracked ? (
-                        <Badge
-                          variant="outline"
-                          className="gap-1 text-[10px] text-emerald-600 dark:text-emerald-400"
-                          render={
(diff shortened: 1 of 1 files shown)


gkhngyk (changes requested):
Nice work — the logic is spot on: pagination (10/page), the numbered controls, the brand-change + post-track resets, the "High frequency" heading, and the intent column all stay intact. 🙌

One thing before merge: the file isn't Prettier-formatted — the wrapped table block needs re-indenting, so CI's `format:check` will fail. Please run `yarn format` from `web/` and commit the result.

Optional / non-blocking: computing `totalPages` and `pageRows` as `const`s just above the `return (` would read a bit cleaner than the inline IIFE — totally your call.

Thanks for the contribution!

gkhngyk (approved):
Thanks for the quick fix — Prettier's happy now (format diff is clean) and the pagination logic is unchanged: 10/page, numbered controls, brand-change + post-track resets, the "High frequency" heading, and the intent column all intact. LGTM! 🎉

PR #357, opened 2026-07-06
fix(prompts): refresh all prompts after tracking fanout query

### LINKED ISSUE #345: fix(prompts): tracking a Query Fan-out sub-query doesn't refresh the All Prompts list
## Problem

On the Prompts page, clicking **+** (Track) on a row in the **Query Fan-out** tab does add the sub-query as a tracked prompt — but the **All Prompts** tab still shows the old list. A user who switches to All Prompts to confirm the addition doesn't see the new prompt and concludes it wasn't added. The prompt *is* there (a page reload or brand switch reveals it); only the in-memory list is stale.

## Root cause (from the code)

The Prompts page is a client component that loads the All Prompts data once into local state, and the Query Fan-out tab refreshes only *its own* data after a track — the two never talk to each other.

- The page holds `allPrompts` in `useState` and populates it in `loadData()`; every other mutating action on the page (`analyze`, `refresh`, etc.) calls `await loadData()` afterwards to re-sync. See [`web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/page.tsx) — `allPrompts` / `setAllPrompts`, `loadData`, and the `<QueryFanoutTab brandId={activeBrandId} />` render (which is passed **no** refresh callback).
- The fan-out tab's track handler adds the prompt and then reloads only the fan-out aggregation (so its own row flips to **Tracked ✓**), but it has no way to tell the page to refresh `allPrompts`. See [`web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/_fanout-tab.tsx) — `handleTrack` → `trackFanoutQuery` → `await load()`.
- `trackFanoutQuery` calls `revalidatePath('/dashboard/prompts')` ([`web/src/lib/actions/fanout.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/fanout.ts)), but the page keeps its list in client `useState`, so path revalidation doesn't re-sync that in-memory state without a navigation/refresh.

## Proposed change

Give the fan-out tab a way to trigger the page's existing refresh so the All Prompts list reflects the new prompt immediately:

- Add an `onTracked?: () => void` (or `onTracked?: () => Promise<void>`) prop to `QueryFanoutTab`, called after a successful `trackFanoutQuery`.
- Wire it at the render site to the page's `loadData` (mirrors how the analyze/refresh actions already re-sync): `<QueryFanoutTab brandId={activeBrandId} onTracked={loadData} />`.

That's the whole fix — no new data flow, just reusing the page's `loadData`. (Optionally, only re-fetch the prompt list rather than all page data if a lighter refresh is preferred.)

## How to test

1. Prompts → **Query Fan-out**, click **+** on an untracked sub-query → toast confirms, the row flips to **Tracked ✓**.
2. Switch to **All Prompts** *without reloading the page* → the newly tracked sub-query now appears in the list (previously it did not until a manual reload).
3. The KPI/"unanalyzed keywords" counts that derive from `allPrompts` also reflect the new prompt.

## Notes

- Small, self-contained UI wiring — good first issue.

AyushSrivastava1818:
## Summary

Refresh the parent Prompts page after successfully tracking a Query Fan-out sub-query.

This PR adds an optional `onTracked` callback to `QueryFanoutTab` and invokes it after a successful `trackFanoutQuery` operation. The parent `Prompts` page passes its existing `loadData` function as the callback, ensuring the **All Prompts** list and related metrics are refreshed immediately without requiring a manual page reload.

## Related issue

Closes #345

## Type of change

- [x] `fix` — Bug fix
- [ ] `feat` — New feature

## Checklist

- [x] `yarn lint` passes (run from `web/`) *(warnings only, no errors)*
- [ ] `yarn typecheck` passes (requires project environment setup)

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
@@ -30,7 +30,12 @@ function platformLabel(slug: string): string {
   return PLATFORM_LABELS[slug] ?? slug;
 }
 
-export function QueryFanoutTab({ brandId }: { brandId: string }) {
+type QueryFanoutTabProps = {
+  brandId: string;
+  onTracked?: () => void | Promise<void>;
+};
+
+export function QueryFanoutTab({ brandId, onTracked }: QueryFanoutTabProps) {
   const [data, setData] = useState<QueryFanoutData | null>(null);
   const [loading, setLoading] = useState(true);
   const [addingKey, setAddingKey] = useState<string | null>(null);
@@ -59,7 +64,6 @@ export function QueryFanoutTab({ brandId }: { brandId: string }) {
       setLoading(false);
     }
   }, [brandId]);
-
   useEffect(() => {
     load();
   }, [load]);
@@ -73,8 +77,9 @@ export function QueryFanoutTab({ brandId }: { brandId: string }) {
     try {
       await trackFanoutQuery(brandId, query);
       toast.success('Added as a tracked prompt');
-      setPage(1);
       await load();
+      setPage(1);
+      await onTracked?.();
     } catch (err) {
       toast.error(err instanceof Error ? err.message : 'Failed to track this query');
     } finally {
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
@@ -619,7 +619,7 @@ export default function PromptsPage() {
 
         {/* ─── Query Fan-out tab ───────────────────────────────────────── */}
         <TabsContent value="fanout" className="mt-4">
-          {activeBrandId && <QueryFanoutTab brandId={activeBrandId} />}
+          {activeBrandId && <QueryFanoutTab brandId={activeBrandId} onTracked={loadData} />}
         </TabsContent>
 
         {/* ─── Insights tab ────────────────────────────────────────────── */}


gkhngyk (changes requested):
Nice — this is exactly the right fix for #345 (the `onTracked` → `loadData` wiring, and the parent passing `loadData`). 🙌

One thing before it can merge: PR #355 (the "High frequency" pagination) landed in `_fanout-tab.tsx` after you branched, so this now conflicts with `main` (GitHub shows it as not mergeable). Please rebase onto the latest `main` and resolve `_fanout-tab.tsx`, keeping **both** the new pagination code and your `onTracked` wiring — in `handleTrack` that means keeping `setPage(1)` and adding `await onTracked?.()` after `await load()`.

Once it's rebased and green I'll re-review. Thanks!

AyushSrivastava1818:
Thanks! I've rebased onto the latest main, resolved the conflict, and kept both the pagination change (setPage(1)) and the onTracked refresh callback as requested. The branch has been force-pushed and is ready for another look.

gkhngyk (approved):
Thanks for the rebase — the conflict's resolved cleanly (`handleTrack` keeps both `setPage(1)` from the pagination work and the new `await onTracked?.()`), it's mergeable, and Prettier's happy. The All Prompts list now refreshes right after tracking a fan-out sub-query. LGTM! 🎉

PR #358, opened 2026-07-06
feat(prompts): Query Fan-out — By prompt grouped view with expandable…

### LINKED ISSUE #350: feat(prompts): "Query Fanouts by Prompts" — group the fan-out by prompt with expandable rows
## Summary

Add a second view to the Query Fan-out tab — **Query Fanouts by Prompts** — that shows the same observed fan-out data, but **grouped by the prompt it came from** instead of one flat list. At the top is a table of the brand's prompts; clicking the expand arrow on a prompt reveals the sub-queries that prompt's answers actually ran.

It's the same data as the "High frequency" list (#349), just pivoted: instead of "sub-query → which prompts sourced it", this is "prompt → which sub-queries it fanned out into".

## What to change

- A **Query Fanouts by Prompts** table: one row per prompt (that has any observed fan-out), showing the prompt text and a count of its distinct sub-queries.
- Each row is **expandable** via a chevron/arrow; expanding reveals that prompt's sub-queries, reusing the same columns as the flat table (sub-query, engine chip, times searched, +/Tracked).
- Empty state when a brand has no fan-out yet (same clear notice as the flat view).

## Where

- [`web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/_fanout-tab.tsx) — the grouped/expandable table.
- [`web/src/lib/actions/fanout.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/fanout.ts) — the data can be **derived from the existing `getQueryFanout` result** (each `subQuery` already carries its `sourcedPrompts`, so invert sub-query→prompts into prompt→sub-queries client-side), or add a dedicated `getQueryFanoutByPrompt(brandId, { days })` action if a server-shaped grouping is cleaner. Either way, no new columns — the sub-query rows reuse the existing shape.

## Layout

- Recommend a toggle on the Query Fan-out tab between **High frequency** (#349, the flat paginated list) and **By prompt** (this view), so the two lenses share the tab.

## Notes

- Depends on the Query Fan-out feature (#342 / #333) and pairs with #349.
- The "+"/Tracked action inside an expanded prompt behaves exactly like the flat list (promote a sub-query to a tracked prompt).

## How to test

1. `/dashboard/prompts?tab=fanout` → **By prompt** view: a table of prompts that produced fan-out, each with its sub-query count.
2. Click a prompt's expand arrow → its observed sub-queries appear (engine, times searched, +/Tracked), matching what the flat "High frequency" list shows for the same sub-queries.
3. A prompt with no fan-out isn't listed; a brand with no fan-out shows the empty state.
4. Tracking a sub-query from inside an expanded prompt works the same as from the flat list.

d180:
## Summary

   On the Prompts page → Query Fan-out tab, the view showed a single flat list of sub-queries with no way to see which tracked prompts drove the most fan-out. Added a **By prompt** toggle that pivots the same data — one collapsible row per prompt, expanding to reveal the sub-queries that prompt's answers actually triggered. The "High frequency" flat list is unchanged; the two lenses share the tab via a segmented toggle.

   Data is derived client-side by inverting the existing `sourcedPrompts` array on each `FanoutSubQuery` — no new server action or DB columns required.

   Also adds fan-out seed data to `supabase/seed.sql` so the tab is populated on a local `db reset`.

   ## Related issue
   
   Closes #350 

   ## Type of change

   - [x] `feat` — New feature

   ## Checklist

### DIFF
--- a/supabase/seed.sql
+++ b/supabase/seed.sql
@@ -528,3 +528,86 @@ INSERT INTO public.prompt_result_shopping_cards (
 )
 ON CONFLICT (id) DO NOTHING;
 
+-- ─── Query Fan-out seed data ──────────────────────────────────────────────────
+-- Populates search_queries on copilot-web and perplexity-web prompt_results so
+-- the Query Fan-out tab has data to display locally. Each item mirrors the real
+-- shape: { query, engine?, source_platform }.
+
+-- Prompt 801: "What are the best coffee subscription services in 2026?"
+UPDATE public.prompt_results
+SET search_queries = '[
+  {"query": "best coffee subscription services 2026", "source_platform": "copilot-web"},
+  {"query": "coffee subscription box comparison", "source_platform": "copilot-web"},
+  {"query": "top rated coffee subscription 2026", "source_platform": "copilot-web"}
+]'::jsonb
+WHERE prompt_id = '88888888-8888-8888-8888-888888888801' AND platform = 'copilot-web';
+
+UPDATE public.prompt_results
+SET search_queries = '[
+  {"query": "best coffee subscription services 2026", "engine": "web", "source_platform": "perplexity-web"},
+  {"query": "top rated monthly coffee boxes", "engine": "web", "source_platform": "perplexity-web"}
+]'::jsonb
+WHERE prompt_id = '88888888-8888-8888-8888-888888888801' AND platform = 'perplexity-web';
+
+-- Prompt 802: "Which specialty coffee subscriptions ship internationally?"
+UPDATE public.prompt_results
+SET search_queries = '[
+  {"query": "international coffee subscription shipping", "source_platform": "copilot-web"},
+  {"query": "specialty coffee worldwide delivery", "source_platform": "copilot-web"}
+]'::jsonb
+WHERE prompt_id = '88888888-8888-8888-8888-888888888802' AND platform = 'copilot-web';
+
+UPDATE public.prompt_results
+SET search_queries = '[
+  {"query": "international coffee subscription shipping", "engine": "web", "source_platform": "perplexity-web"},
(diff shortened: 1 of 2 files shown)


gkhngyk (changes requested):
Really clean work — the shared components (`TrackCell` / `EngineBadges` / `IntentBadge`), the `useMemo` inversion for the **By prompt** view, the expandable rows, keeping the pagination + `setPage(1)`, and the `seed.sql` fan-out data all look great. 🎉

One fix before merge: in `ByPromptView`, `groups.map(...)` returns a bare `<>` fragment with no key, which triggers React's "unique key" warning (the inner `TableRow` keys don't satisfy the map's root element). Please switch it to `<Fragment key={prompt.id}>…</Fragment>` (import `Fragment` from `react`).

Everything else LGTM — ping me once that's in and I'll re-review.

Heads-up on sequencing: this touches `_fanout-tab.tsx` (same as PR #357, still open), so whichever merges second will need a small rebase.

d180:
Hi @gkhngyk , I've rebased the branch on top of main and resolved the conflicts. Also applied the fix you suggested — replaced <> with <Fragment key={prompt.id}> in ByPromptView's groups.map() so the key is now on the root element. Please review and let me know if anything needs attention!

gkhngyk (approved):
Thanks — the Fragment key fix is in (`<Fragment key={prompt.id}>`, no more React key warning). I re-ran everything in an isolated worktree: typecheck, Prettier, and the fanout tests all pass, and the rebase correctly keeps the pagination (#355), the intent column (#353), and the `onTracked` wiring (#357) while adding the By-prompt grouped/expandable view. Clean work — LGTM! 🎉

PR #367, opened 2026-07-07
fix(content): suppress skeleton flash on filter change and bulk actions

### LINKED ISSUE #123: ci: run ESLint for web in CI
## Problem

The web package defines a `lint` script (`web/package.json`) but the CI workflow never runs it. Today `.github/workflows/ci.yml` only runs Prettier (`format:check`) and `typecheck` for `web`:

```yaml
      - name: Prettier format check
        run: yarn format:check
      - name: TypeScript typecheck
        run: yarn typecheck
```

This means ESLint violations can land on `main` without being caught.

## Proposed change

Add an ESLint step to the existing `web` job in `.github/workflows/ci.yml`:

```yaml
      - name: ESLint
        run: yarn lint
```

## Acceptance criteria

- [x] `yarn lint` runs on every PR and on push to `main`
- [x] The repo currently passes `yarn lint` (fix or triage any existing violations first)
- [x] Job fails the build on lint errors

## Notes

Good first issue — single workflow file change plus verifying the existing codebase is lint-clean.

### LINKED ISSUE #365: fix(content): the list flashes its skeleton on every filter change and bulk action
## Problem

On the Content Optimization page (`/dashboard/content`), the whole opportunity list flashes to its loading **skeleton** on almost every interaction — changing a filter (Status / Impact / Type) or running a bulk action (**Done** / **Dismiss all** / **Send all**) briefly wipes the list and re-renders skeletons. It's jarring and makes the page feel like it fully reloads each time. The skeleton should only show on the **initial** load; subsequent refreshes should update the list in place.

## Root cause (from the code)

In [`web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/content/page.tsx):

- `loadData` always starts with `setLoading(true)`, and the list render is gated on that `loading` flag (skeletons while true).
- `loadData` is a `useCallback` whose deps include `statusFilter`, `impactFilter`, `typeFilter`, so **changing any filter recreates it**, and the `useEffect(() => { loadData() }, [loadData])` re-runs it — full skeleton every filter change.
- The bulk handlers (`handleBulkDone`, `handleBulkDismiss`, `handleBulkSend`) each `await loadData()` afterward — full skeleton after every bulk action too.

So every filter change and every bulk action toggles the full-page skeleton.

## Proposed change

Distinguish the **initial** load from **re-fetches**, so only the first load shows the skeleton:

- Add a `{ silent }` option to `loadData` that skips `setLoading(true)` (mirrors the pattern the Insights page already uses: `loadData(overrideFilters, { silent })`).
- Call the silent variant for filter-change refreshes and after bulk actions, keeping the current list visible while it updates (optionally a subtle inline/“refreshing” indicator instead of the skeleton).

Alternatively (simpler), only render the skeleton when the list is actually empty — e.g. gate it on `loading && opportunities.length === 0` — so a populated list stays put during a refetch.

## How to test

1. Load `/dashboard/content` → the skeleton shows once on first load.
2. Change the Status / Impact / Type filter → the list updates **without** flashing back to the skeleton.
3. Select some opportunities and click **Done** / **Dismiss all** / **Send all** → the list updates in place, no skeleton flash.
4. Switching brand (a genuinely new dataset) may still show the skeleton — that's fine.

## Notes

- Same class of issue as #346 (the Query Fan-out track skeleton flash) — reusing the initial-vs-refetch distinction.
- Small, self-contained UI change.

Maqbool61:
Closes #365

- Added silent parameter (default false) to loadData
- When silent=true, skips setLoading(true) so existing rows stay visible during background refresh
- Initial load and filter-change calls remain non-silent (correct full loading state on first paint and filter switch)
- All post-mutation calls use loadData(true): handleSendWebhook, handleDismiss, handleBulkSend, handleBulkDone, handleBulkDismiss, and pollJob completion handler
- Zero new TypeScript errors introduced (verified via tsc --noEmit diff)

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

- [ ] Branch follows the naming convention (`feature/`, `fix/`, `chore/`, `docs/`) — see [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md)
- [ ] Commits follow [Conventional Commits](https://www.conventionalcommits.org/)
- [ ] Changes are focused — one concern per PR

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
@@ -151,38 +151,41 @@ export default function ContentPage() {
   const [webhookOpen, setWebhookOpen] = useState(false);
   const pollRef = useRef(false);
 
-  const loadData = useCallback(async () => {
-    if (!activeBrandId) {
-      setOpportunities([]);
-      setTotal(0);
-      setLoading(false);
-      return;
-    }
+  const loadData = useCallback(
+    async (silent = false) => {
+      if (!activeBrandId) {
+        setOpportunities([]);
+        setTotal(0);
+        setLoading(false);
+        return;
+      }
 
-    setLoading(true);
-    setSelectedIds(new Set());
-    try {
-      const filters: Record<string, string> = {};
-      if (statusFilter !== 'all') filters.status = statusFilter;
-      if (impactFilter !== 'all') filters.impact = impactFilter;
-      if (typeFilter !== 'all') filters.type = typeFilter;
+      if (!silent) setLoading(true);
+      setSelectedIds(new Set());
+      try {
+        const filters: Record<string, string> = {};
+        if (statusFilter !== 'all') filters.status = statusFilter;
+        if (impactFilter !== 'all') filters.impact = impactFilter;
+        if (typeFilter !== 'all') filters.type = typeFilter;
 
-      const data = await getOpportunities(activeBrandId, {
-        ...filters,
-        limit: 100,
-        sort: 'score',
-      });
-      setOpportunities(data.opportunities);
-      setTotal(data.total);
-      return data.total;
-    } catch (err) {
-      console.error('Failed to load opportunities:', err);
-      toast.error('Failed to load content opportunities');
-      return 0;
-    } finally {
-      setLoading(false);
-    }
-  }, [activeBrandId, statusFilter, impactFilter, typeFilter]);
+        const data = await getOpportunities(activeBrandId, {
+          ...filters,
+          limit: 100,
+          sort: 'score',
+        });
+        setOpportunities(data.opportunities);
+        setTotal(data.total);
+        return data.total;
+      } catch (err) {
+        console.error('Failed to load opportunities:', err);
+        toast.error('Failed to load content opportunities');
+        return 0;
+      } finally {
+        setLoading(false);
+      }
+    },
+    [activeBrandId, statusFilter, impactFilter, typeFilter],
+  );
 
   useEffect(() => {
     loadData();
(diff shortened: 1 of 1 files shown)


gkhngyk (changes requested):
Nice — the bulk-action skeleton flash is fixed with the `silent` param (`loadData(true)` on the mutation paths). 🙌

One thing before it fully closes #365: the issue also covers **filter changes** (the reporter specifically hit "selecting a filter option runs the skeleton"), and that path still flashes — the `useEffect(() => loadData(), [loadData])` re-runs **non-silent** whenever a Status/Impact/Type filter changes (its deps include the filters), so `setLoading(true)` fires and the list drops to the skeleton.

Could you suppress that case too? Two clean options:
- (a) make the filter-change effect call `loadData(true)` after the first load — distinguish the initial mount via a `hasLoadedRef`; or
- (b) simpler — gate the skeleton render on `loading && opportunities.length === 0`, so a populated list never flashes during any refetch (filter change *or* mutation), only the first empty load does.

Either one makes it satisfy the filter-change step (step 2 of the How-to-test in the issue). Thanks!

gkhngyk (approved):
Perfect — gating the skeleton on `loading && opportunities.length === 0` covers both cases cleanly: a filter change keeps the populated list in place (no flash), and the initial empty load still shows the skeleton. Combined with `loadData(true)` on the mutation paths, both the filter-change and bulk-action flashes from #365 are gone. Prettier's happy. LGTM! 🎉

PR #369, opened 2026-07-07
fix(content): generate opportunities in brand language

### LINKED ISSUE #363: fix(content): opportunities are generated in English regardless of the brand's language
## Problem

When a brand's language is set to something other than English (e.g. Turkish, chosen during onboarding), the **Content Optimization** opportunities (`/dashboard/content`) are still generated in **English**. They should be written in the brand's language.

## Root cause (from the code)

The content-opportunity generator never tells the LLM which language to write in — unlike the other AI-generation flows, which already do.

In [`server/src/workers/content-worker.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/workers/content-worker.js), `processContentJob`:

- fetches the brand with `select('id, name, description, industry, organization_id')` — **`language` is not selected**;
- `OPPORTUNITY_SYSTEM_PROMPT` and the built `userPrompt` contain **no language instruction**;
- so `generateObject({ system: OPPORTUNITY_SYSTEM_PROMPT, prompt: userPrompt, … })` defaults to English.

For contrast, the flows that do it right: the prompt-suggestion generator resolves `getLanguageName(brand.language)` and instructs the model to write in that language (see `loadSuggestionContext` / `generateSuggestions` in [`server/src/routes/prompts.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/prompts.js)), and the topic/competitor suggestion routes append e.g. `IMPORTANT: All … MUST be written in ${langName}`.

## Proposed change

Mirror the existing pattern in `content-worker.js`:

1. Add `language` to the brand `select(...)`.
2. `import { getLanguageName } from '../lib/languages.js';` and resolve `const langName = getLanguageName(brand.language);`.
3. Add a language directive to the generation prompt, e.g. append to the system prompt or the user prompt: `IMPORTANT: Write every opportunity title and description in ${langName}.`

The opportunity `type` / `impact` enums stay as-is (they're internal values, not shown-as-text) — only the human-readable `title` and `description` need to be in the brand's language.

## How to test

1. Create/set a brand whose language is **Turkish** (or any non-English), then generate Content Optimization opportunities.
2. The generated **titles and descriptions** come back in that language (Turkish), not English.
3. An English brand still gets English opportunities (no regression).

## Notes

- Server-only change in the content generation worker; no schema or web change.
- Uses the existing `getLanguageName` helper and the same convention as prompt/topic/competitor suggestions.

Sam-syntax35:
## Summary

Fixes content opportunity generation so it respects the brand's configured language instead of always generating English.

## Changes

- Select `language` when fetching the brand.
- Import and use `getLanguageName()` to resolve the language name.
- Pass a language instruction to the LLM prompt so opportunity titles and descriptions are generated in the brand's language.
- Leave internal enum values (`type`, `impact`) unchanged.

Fixes #363

### DIFF
--- a/server/src/workers/content-worker.js
+++ b/server/src/workers/content-worker.js
@@ -6,6 +6,7 @@
 import { generateObject } from 'ai';
 import { z } from 'zod';
 import { resolveModel } from '../lib/ai-provider.js';
+import { getLanguageName } from '../lib/languages.js';
 import supabaseAdmin from '../config/supabase.js';
 import logger from '../lib/logger.js';
 
@@ -82,11 +83,12 @@ export async function processContentJob({ brandId, model, job }) {
 
   const { data: brand } = await supabaseAdmin
     .from('brands')
-    .select('id, name, description, industry, organization_id')
+    .select('id, name, description, industry, organization_id, language')
     .eq('id', brandId)
     .single();
 
   if (!brand) throw new Error(`Brand not found: ${brandId}`);
+  const langName = getLanguageName(brand.language);
 
   const { data: domains } = await supabaseAdmin
     .from('brand_domains')
@@ -215,7 +217,9 @@ ${scoredPrompts
   )
   .join('\n')}
 
-Generate actionable content opportunities based on this data. Reference specific numbers and competitors where relevant.`;
+Generate actionable content opportunities based on this data. Reference specific numbers and competitors where relevant.
+
+IMPORTANT: Write every opportunity title and description in ${langName}.`;
 
   const aiModel = resolveModel(model);
 


gkhngyk (approved):
Clean fix — matches the issue exactly: `language` added to the brand select, `getLanguageName` resolves it (falls back to English for an unset language, so no regression), and the `IMPORTANT: Write … in ${langName}` directive is appended to the prompt, mirroring the prompt / topic-suggestion flows. The `type`/`impact` enums are left untouched, and Prettier's happy. LGTM! 🎉

Worth a quick manual sanity-check generating opportunities on a Turkish (or other non-English) brand, but the pattern's proven.

PR #370, opened 2026-07-07
refactor(prompts): use shared compact number formatter

Sam-syntax35:
## Summary

This PR extracts a shared `formatCompactNumber()` helper and uses it across the Prompts dashboard.

### Changes

- Added `formatCompactNumber()` in `web/src/lib/format.ts`
- Replaced inline volume formatting in:
  - Prompt Insights KPI cards
  - Google searches subtext
  - Average AI Volume card
  - VolumePill
  - Volume cluster values
  - Prompt Volumes chart axis
- Replaced the inline formatter in `similar-topics/page.tsx`

### Result

- `1529k` → `1.5M`
- `10195k` → `10.2M`
- `1000k` (chart axis) → `1M`

No changes to the underlying calculations; presentation only.


gkhngyk (changes requested):
Really nice — the shared `formatCompactNumber` (k → M → B with the trailing `.0` trimmed) is used everywhere including the chart axis, and swapping the duplicated `formatVolume` in `similar-topics` for it gives one source of truth. This fixes both `1529k` (the cards) and `1000k` (the chart axis). I ran typecheck in a worktree — passes. 🎉

One thing before merge: Prettier's failing on two files — `web/src/lib/format.ts` is missing its trailing newline, and `similar-topics/page.tsx` has a couple of stray blank lines where `formatVolume` was removed. Just run `yarn format` from `web/` and commit. Then it's good to go!

gkhngyk (approved):
Thanks for the format fix — all four files are Prettier-clean now (`format.ts` and `similar-topics` were the two that were failing). The logic is unchanged from the last pass (typecheck passed in a worktree), and `formatCompactNumber` is used consistently everywhere including the chart axis, replacing the duplicated `formatVolume`. Fixes both `1529k` (cards) and `1000k` (chart axis). LGTM! 🎉

Issue #348, opened 2026-07-06
ux(prompts): volume abbreviations should roll over to M/B — cards show "1529k", chart axis shows "1000k"


gkhngyk:
## Problem

On the Prompts page → **Insights** tab, large volumes are abbreviated to `k` even past a million, which reads badly in two places:

**1. The KPI stat cards:**
> **Total Est. AI Volume**
> ~1529k
> from 10195k Google searches

`~1529k` is 1,529,000 — it should read `~1.5M`, and `10195k` should read `~10.2M`.

**2. The Prompt Volumes chart axis:** the X-axis shows `1000k` (right after `750k`) instead of `1M`.

The formatter divides by 1,000 and always appends `k`, so it never rolls over to `M` (or `B`).

## Root cause (from the code)

Several spots hard-code `(value / 1000).toFixed(n) + 'k'` with no million rollover. In [`web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/page.tsx):

- The **Total Est. AI Volume** stat card: `value={`~${(totalAiVol / 1000).toFixed(0)}k`}` and `sub={`from ${(totalGoogleVol / 1000).toFixed(0)}k Google searches`}`.
- The average stat card: `~${(totalAiVol / volumes.length / 1000).toFixed(1)}k`.
- `VolumePill`: `value >= 1000 ? `~${(value / 1000).toFixed(1)}k` : …` (so a 1.5M pill renders "~1529.0k").
- The volume-cluster row value: `item.volume >= 1000 ? `${(item.volume / 1000).toFixed(1)}k` : …`.

And the **Prompt Volumes chart** (`PlatformVolumeChart`) axis tick in [`_charts.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/_charts.tsx): `tickFormatter={(v) => (v >= 1000 ? `${(v / 1000).toFixed(0)}k` : `${v}`)}` — this is the `1000k` on the axis.

A correct formatter already exists inline in [`prompts/similar-topics/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/similar-topics/page.tsx):

```ts
[...]

PR #375, opened 2026-07-08
Perf/citations memoize classification

Sam-syntax35:
## Summary

Memoize expensive domain and article type classification during citations aggregation.

### Changes
- Cache `classifyDomain()` results by host in `getCitationsOverview()`
- Cache `classifyArticleType()` results by URL + title in `getCitationsOverview()`
- Cache `classifyDomain()` results by host in `getCitationGaps()`

### Why

Previously, repeated hosts and URLs were classified on every citation occurrence. This change memoizes classification results for the duration of each aggregation call, reducing repeated work while preserving identical behavior.

### Testing

- Verified the project builds successfully.
- Verified Citations overview and Competitor Gaps continue to behave as expected.
- No database or schema changes.


gkhngyk (approved):
The citations classification memoization is exactly right — per-call `domainClassificationCache` (by host) + `articleTypeCache` (by `url\ntitle`), looked up before calling `classifyDomain` / `classifyArticleType`, in both `getCitationsOverview` and `getCitationGaps`. Behavior-preserving (`classifyCtx` is constant per call, and a cached `null` is distinguished from a miss via `=== undefined`). I verified typecheck + Prettier in a worktree — both pass. 🎉

Heads-up (harmless): this branch was cut from your #348 branch, so the diff also shows the `formatCompactNumber` files — but those are byte-identical to `main` now that #370 merged, so the **net** change vs `main` is only `citations.ts`. A squash-merge will produce a clean citations-only commit. (If you'd rather have a tidy diff, a quick rebase onto `main` would drop the format files — totally optional.)

PR #387, opened 2026-07-09
fix(web): prevent fanout intent badge pop-in

### LINKED ISSUE #380: ux(prompts): Query Fan-out Intent column pops in a few seconds after the table loads
## Problem

On the Prompts page → **Query Fan-out** tab (`/dashboard/prompts?tab=fanout`), the table paints first and the **Intent** column stays empty (a `—` placeholder) for a few seconds, then the intent badges pop in afterward. It reads as if the row loaded broken and got patched a moment later. Expected: the table and its Intent badges appear **together**, as one complete load.

## Root cause (from the code)

In [`web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/_fanout-tab.tsx), intent classification is intentionally **non-blocking**:

```ts
const result = await getQueryFanout(brandId, { days: 30 });
setData(result);            // table paints immediately
// non-blocking: badges fill in when this resolves
classifyFanoutIntents(queries)
  .then((map) => setIntents((prev) => ({ ...prev, ...map })))
  .catch(() => {});
setLoading(false);
```

`IntentBadge` renders a `—` placeholder until the row's intent resolves. Since `classifyFanoutIntents` is an LLM classification round-trip, the badges land a few seconds after the table — hence the pop-in. (It's cached server-side, so it's slow only on the first classification of a given set of sub-queries; repeat loads are fast.)

## Proposed change

Fold the intent fetch into the initial load so the table and Intent badges paint together, **with a timeout fallback** so a slow first-time classification never blocks the table indefinitely:

- In `load()`, `await` `classifyFanoutIntents(queries)` (populating `intents`) **before** flipping `setLoading(false)` — but race it against a short timeout (e.g. ~1.5s).
- If intents resolve within the timeout (the common, cached case): the table renders once, fully populated — no pop-in.
- If they don't (rare first-time/slow case): fall back to today's behavior — paint the table now and let the badges fill in when they arrive, so we never hold the whole tab on a skeleton waiting for the LLM.
- Keep the server-side cache and the `—` placeholder as the fallback state.

## Acceptance criteria

- [ ] On a normal (cached) load, the Query Fan-out table and its Intent badges appear at the same time — no visible pop-in.
- [ ] A slow/first-time intent classification does not block the table beyond the timeout; it still degrades to the current fill-in-later behavior.
- [ ] Applies to both the High-frequency view and the By-prompt view (both use `IntentBadge` / the shared `intents` map).
- [ ] No extra network calls vs today (still one `classifyFanoutIntents` batch).

## Notes

- Presentation/timing only — no change to how intents are computed or cached.

d180:
## Summary

  Fixes the Query Fan-out table loading behavior so intent badges no longer
  visibly pop in after the table renders on normal cached loads.

  The fan-out tab now waits up to 1.5s for the existing batched intent
  classification request during the initial load. If intents resolve within that
  window, the table and badges render together. If classification is slow, the
  table still renders after the timeout and the same request fills badges later,
  preserving the existing fallback behavior without adding extra network calls.

  ## Related issue

  Closes #380

  ## Type of change

  - [x] `fix` — Bug fix

  ## Checklist

  - [x] Branch follows the naming convention (`feature/`, `fix/`, `chore/`,
  `docs/`) — see [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/
  main/CONTRIBUTING.md)
  - [x] Commits follow [Conventional Commits](https://
  www.conventionalcommits.org/)

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
@@ -43,6 +43,8 @@ type QueryFanoutTabProps = {
   onTracked?: () => void | Promise<void>;
 };
 
+const INTENT_INITIAL_LOAD_TIMEOUT_MS = 1500;
+
 export function QueryFanoutTab({ brandId, onTracked }: QueryFanoutTabProps) {
   const [data, setData] = useState<QueryFanoutData | null>(null);
   const [loading, setLoading] = useState(true);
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Clean, focused fix — the intent badges and the table now paint together on a warm load, and reusing the same promise means no extra network call. Verified typecheck + prettier pass. Thanks!

Issue #384, opened 2026-07-09
i18n: wire the remaining dashboard pages to next-intl (Insights, Topics, Agent)


gkhngyk:
## Summary

Most of the dashboard already localizes its UI via `next-intl` (`useTranslations(...)` + `web/messages/en.json` namespaces — e.g. `audit`, `reports`). A few pages still render **hardcoded English** and never call `useTranslations`. This is a consistency / infrastructure tracking issue to bring them in line with the rest of the dashboard.

> **Scope note:** `web/messages/` currently ships **only `en.json`** (no second locale yet), so this is a **consistency + infra** cleanup, not an immediately visible translation win. It makes every dashboard page follow the same `t('…')` pattern so a future locale drop-in "just works". Each page below can be its own small PR.

### 1. Insights — `web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx`
- No `useTranslations` import; all text hardcoded.
- An **`insights` namespace already exists** in `web/messages/en.json` (≈14 keys: `title`, `description`, `runPrompts`, `promptResults`, `platform`, `mentions`, `citations`, `sentiment`, `score`, `positive`, `neutral`, `negative`) but is **unused**.
- [ ] Wire the page to the existing `insights` namespace.
- Note: the namespace only covers ~14 strings; the page has far more text without keys. Wiring the existing keys is in scope; **adding keys for the rest is a follow-up** (out of scope here).

### 2. Topics — `web/src/app/[locale]/(dashboard)/dashboard/topics/page.tsx`
- No `useTranslations`; header (~265–268), KPI labels/subs (~305–335), and empty state (~361–366) are hardcoded.
- **No `topics` page namespace exists yet** (`"topics"` in `en.json` is the nav label). Must **add** a page namespace and wire the page — follow the existing `audit` / `reports` pattern.
- [ ] Add a `topics` page namespace to `en.json` + wire the page.

### 3. Agent — `web/src/app/[locale]/(dashboard)/dashboard/agent/page.tsx`
- No `useTranslations`; hardcoded strings include `confirm('Delete this conversation?')` (~248), "New chat" (~292), "No conversations yet." (~296), the input placeholder (~373), "Add your Anthropic API key" / "Go to Settings" (~674–685), plus `toast.success('Copied')` and aria-labels.
- **No `agent` page namespace** (`"agent"` in `en.json` is the nav label). Must **add** an `agent` namespace and wire the page.
[...]

PR #389, opened 2026-07-10
i18n: wire insights page to next-intl

Sam-syntax35:
## Summary

Wired the Insights dashboard page to `next-intl` by:

- importing `useTranslations`
- initializing the `insights` translation namespace
- replacing the page title with the existing `insights.title` translation key

This keeps the existing English copy unchanged while aligning the page with the dashboard's localization pattern.

Part of #384.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx
@@ -1606,7 +1606,7 @@ export default function InsightsPage() {
     return (
       <div className="space-y-6">
         <div>
-          <h1 className="text-2xl font-bold tracking-tight">Answer Engine Insights</h1>
+          <h1 className="text-2xl font-bold tracking-tight">{t('title')}</h1>
           <p className="text-muted-foreground text-sm">{brand.name}</p>
         </div>
         <TrackingProgressBanner jobStatus={jobStatus} onStop={handleStopTracking} />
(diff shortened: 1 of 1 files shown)


gkhngyk (changes requested):
Thanks — the title key resolves correctly and the copy is unchanged, so no risk there. A couple of things before this is ready:

1. **The same `<h1>` appears twice.** The empty-state branch (`if (trulyEmpty)`, ~line 1607) still has a hardcoded `Answer Engine Insights`, while the main render (~1645) now uses `t('title')`. Please apply `t('title')` there too so the title is consistent — otherwise a brand with no data still sees the hardcoded string.
2. **`Closes #384` will close the whole tracking issue on merge**, but Topics and Agent are still open items there. Please change it to `Part of #384` (the parenthetical "(Insights portion)" doesn't stop GitHub from auto-closing).

Optional: wiring the other header strings ("Last run:", the description) in the same pass would be a bonus, but keeping this focused on the title is fine as long as it's consistent.

Once the empty-state title and the issue reference are updated, happy to approve.

Sam-syntax35:
I've updated the empty-state title to use `t("title")` and changed the PR description to `Part of #384`. Please let me know if anything else needs adjusting.

gkhngyk (approved):
Both asks addressed — the empty-state `<h1>` now uses `t('title')` so the title is consistent in both branches, and the issue reference is now `Part of #384` so the tracking issue stays open for the Topics/Agent items. CI is green. Thanks!

gkhngyk (approved):

PR #397, opened 2026-07-10
fix(shopping): display labels instead of raw select values

### LINKED ISSUE #396: bug: shopping filter selects show raw value instead of display label (e.g. "30d" instead of "Last 30d")
## Description

On the Shopping page (`/dashboard/shopping`), the three filter selects in the filter bar — **Date range**, **Platform**, and **Region** — show the raw option *value* in the closed trigger instead of the option's display label. For example, after picking "Last 30d" the trigger shows **"30d"**; the date range default shows **"all"** instead of **"All time"**, and the platform/region selects show **"all"** instead of **"All platforms"** / **"All regions"**.

All three are the same bug in one component, hence a single issue.

## Steps to reproduce

1. Open **Shopping** (`/dashboard/shopping`) for a brand with shopping mode enabled.
2. Look at the filter bar at the top (Date range / Platform / Region).
3. Open the date range select and pick "Last 30d".
4. The dropdown items are labeled correctly, but the closed trigger shows the raw value ("30d", "all", …).

## Expected behavior

The trigger shows the same label as the selected dropdown item: "Last 30d", "All time", "All platforms", "All regions".

## Actual behavior

The trigger shows the raw value: "30d", "all".

## Deployment mode

Cloud (ansvisor.com) — reproduces on self-hosted too; it's purely a frontend rendering issue.

---

## Root cause & suggested fix (for contributors)

The shared select component (`web/src/components/ui/select.tsx`) is built on **Base UI** (`@base-ui/react/select`), not Radix. Unlike Radix — where `<SelectValue />` echoes the selected item's rendered label — Base UI's `<Select.Value />` renders the raw value string when given no children.

The Shopping filter bar (`FilterBar` in `web/src/app/[locale]/(dashboard)/dashboard/shopping/page.tsx`, around lines 770–816) uses a bare `<SelectValue />` in all three selects, so the trigger displays the value instead of the label.

**Suggested approach** — give each `SelectValue` a children function that maps value → label, mirroring the label logic already used for the `SelectItem`s. For example, for the date range select:

```tsx
<SelectValue>
  {(v: string) => (v === 'all' ? 'All time' : `Last ${v}`)}
</SelectValue>
```

and for platform/region:

```tsx
<SelectValue>{(v: string) => (v === 'all' ? 'All platforms' : v)}</SelectValue>
<SelectValue>{(v: string) => (v === 'all' ? 'All regions' : v)}</SelectValue>
```

(Alternatively, Base UI's `items` prop on `Select.Root` can provide the value→label mapping — either approach is fine, just keep it consistent for all three selects.)

**How to verify:** open `/dashboard/shopping`, change each of the three filters, and confirm the closed trigger always shows the human-readable label. Consider grepping for other bare `<SelectValue />` usages backed by value≠label options — but keep this PR focused on the Shopping filter bar.

Please follow [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for branch naming and Conventional Commits.

Sam-syntax35:
## Summary

Fixes the Shopping filter selects so the closed trigger displays the human-readable label instead of the raw value.

### Changes

* Updated the **Date range** select to display labels such as:

  * `All time` instead of `all`
  * `Last 30d` instead of `30d`
* Updated the **Platform** select to display `All platforms` instead of `all`.
* Updated the **Region** select to display `All regions` instead of `all`.
* Kept the existing dropdown options and behavior unchanged.

### Testing

* Verified the Shopping page (`/dashboard/shopping`).
* Confirmed that selecting different values updates the closed trigger with the correct display label.
* Confirmed the dropdown items continue to render and function as expected.

Closes #396.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/shopping/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/shopping/page.tsx
@@ -87,6 +87,8 @@ import {
   getAIProviderDisplayName,
 } from '@/components/ai-provider-avatar';
 import { toCsv } from '@/lib/csv';
+import { PLATFORM_LABELS } from '@/config/platform-labels';
+import { formatRegionDisplay } from '@/lib/region';
 
 const DATE_PRESETS: ShoppingDatePreset[] = ['7d', '30d', '90d', 'all'];
 
@@ -771,7 +773,7 @@ function FilterBar({
       <FilterField label={t('filterDateRange')}>
         <Select value={datePreset} onValueChange={(v) => setDatePreset(v as ShoppingDatePreset)}>
           <SelectTrigger className="h-8 w-32">
-            <SelectValue />
+            <SelectValue>{(value) => (value === 'all' ? 'All time' : `Last ${value}`)}</SelectValue>
           </SelectTrigger>
           <SelectContent>
             {DATE_PRESETS.map((p) => (
@@ -786,13 +788,19 @@ function FilterBar({
       <FilterField label={t('filterPlatform')}>
         <Select value={platform} onValueChange={(v) => setPlatform(v ?? 'all')}>
           <SelectTrigger className="h-8 w-40">
-            <SelectValue />
+            <SelectValue>
+              {(value) =>
+                value === 'all' || !value
+                  ? 'All platforms'
+                  : (PLATFORM_LABELS[String(value)] ?? String(value))
+              }
+            </SelectValue>
           </SelectTrigger>
           <SelectContent>
             <SelectItem value="all">All platforms</SelectItem>
             {platformOpts.map((p) => (
               <SelectItem key={p} value={p}>
-                {p}
+                {PLATFORM_LABELS[p] ?? p}
               </SelectItem>
             ))}
           </SelectContent>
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Thanks for the clean fix — and welcome! This goes a step beyond the issue's suggestion in a good way: reusing the canonical `PLATFORM_LABELS` map and the `formatRegionDisplay` helper keeps the trigger and the dropdown items consistent with the rest of the app.

One small, non-blocking suggestion: the value→label logic is now duplicated between each trigger's `SelectValue` children function and the corresponding `SelectItem` rendering (e.g. `PLATFORM_LABELS[p] ?? p` appears in both). If one side changes later, the other is easy to miss. Extracting tiny local helpers in `FilterBar` would remove that drift risk:

```tsx
const dateLabel = (v: string) => (v === 'all' ? 'All time' : `Last ${v}`);
const platformLabel = (v: string) => (v === 'all' ? 'All platforms' : (PLATFORM_LABELS[v] ?? v));
const regionOptionLabel = (v: string) => (v === 'all' ? 'All regions' : formatRegionDisplay(v));
```

…and using them in both the trigger and the items. Feel free to do it here or leave it as-is — approving either way.

PR #400, opened 2026-07-10
fix(content): display user friendly error when sending with no webhoo…

Sam-syntax35:
## Summary

Fixes the production-only error handling for **"Send to Workflow"** when no workflow is connected.

### Changes

* Updated `sendToWebhook` to return a structured success/error result instead of throwing for expected failures.
* Updated both client call sites to handle the returned result and display a user-friendly toast.
* Preserved fallback error handling for unexpected failures.

### Result

Users now see a concise error message (e.g. **"No workflow connected"**) instead of the generic Next.js Server Components error in production.

### DIFF
--- a/server/src/routes/content.js
+++ b/server/src/routes/content.js
@@ -238,9 +238,7 @@ router.post('/bulk/send-webhook', async (req, res) => {
       .single();
 
     if (!webhookConfig) {
-      return res
-        .status(400)
-        .json({ error: 'No active webhook configured for this brand. Set up a webhook first.' });
+      return res.status(400).json({ error: 'No workflow connected' });
     }
 
     const { data: brand } = await supabaseAdmin
@@ -394,9 +392,7 @@ router.post('/:id/send-webhook', async (req, res) => {
       .single();
 
     if (!webhookConfig) {
-      return res
-        .status(400)
-        .json({ error: 'No active webhook configured for this brand. Set up a webhook first.' });
+      return res.status(400).json({ error: 'No workflow connected' });
     }
 
     const { data: brand } = await supabaseAdmin
--- a/web/src/app/[locale]/(dashboard)/dashboard/content/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/content/[id]/page.tsx
@@ -117,13 +117,17 @@ export default function ContentDetailPage() {
   const handleSend = async () => {
     setSending(true);
     try {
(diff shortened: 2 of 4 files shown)


gkhngyk (changes requested):
Thanks for picking this up — the approach is exactly right: returning a discriminated result from the Server Action instead of throwing is the correct way around Next.js's production error redaction, and both single-send call sites now surface the real message. Server messages are nicely shortened too.

One change I'd like before merging, and one optional nit:

### 1. Bulk send still has the same bug (please fix in this PR)

`bulkSendToWebhook` in `web/src/lib/actions/content.ts` still `throw`s on `!res.ok`, and `handleBulkSend` in `web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx` toasts `err.message`. So a user who selects multiple opportunities and clicks bulk **Send to Workflow** with no workflow connected will still see the long generic "An error occurred in the Server Components render…" toast in production — the exact bug this PR fixes for single send.

Suggested fix, mirroring what you did for `sendToWebhook`:

```ts
export type BulkSendResult =
  | { success: true; sent: number; failed: number }
  | { success: false; error: string };

export async function bulkSendToWebhook(ids: string[]): Promise<BulkSendResult> {
  // ...
  if (!res.ok) {
    const body = await res.json().catch(() => ({}));
    return { success: false, error: body.error || `Server error: ${res.status}` };
  }
  const data = await res.json();
  return { success: true, sent: data.sent, failed: data.failed };
}
```

…and in `handleBulkSend`, toast `result.error` when `success` is false (keeping the `catch` as a short generic fallback).

### 2. Nit (optional)

In `sendToWebhook`, the client-side remap:

```ts
if (error.includes('No active webhook configured') || error.includes('No active webhook')) {
```

the first condition is redundant — the second already covers it. If the intent is backward compatibility with older self-hosted servers that still return the long message, a one-line comment saying so would help; otherwise it can be dropped since the server change in this PR makes it dead code.

Happy to approve once bulk send is covered — everything else looks good and CI is green.

Sam-syntax35:
Addressed the requested changes:

1.Bulk Send now uses the same structured error handling as Single Send.
2.Simplified the redundant webhook error check.
3.Fixed the Prettier formatting issue.

gkhngyk (approved):
Both points addressed — bulk send now uses the same result pattern and the redundant condition is gone. CI is green. Thanks!

PR #402, opened 2026-07-10
fix(prompts): prevent toast after cancelled page navigation

### LINKED ISSUE #394: fix(prompts): navigating away mid-load fires a spurious "Failed to load prompt data" toast
## Problem

Navigating away from **Prompts** (`/dashboard/prompts`) while its data is still loading fires a **"Failed to load prompt data"** error toast — on the page you just navigated *to*.

Repro:
1. Open `/dashboard/prompts` (or switch to it from the sidebar)
2. Quickly click another sidebar item before the page finishes loading
3. The destination page shows the "Failed to load prompt data" toast, even though nothing actually failed

## Cause

`loadData` in [`web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/page.tsx) (~line 308) runs three server actions in parallel and treats **any** rejection as a failure:

```ts
} catch (err) {
  console.error('Failed to load prompt data:', err);
  toast.error('Failed to load prompt data');
}
```

When the route changes mid-flight, the in-flight Server Action fetches are aborted; the rejection lands in this `catch`, and since the sonner toaster lives in the persistent layout, the toast surfaces on the next page. There's no cancellation guard distinguishing "component unmounted / navigation aborted" from a real failure.

## Fix

Use the cancellation-flag pattern that already exists elsewhere in the codebase (e.g. the report detail page):

- Track a `cancelled` flag in the effect (set in the cleanup) and skip **both** the state updates and the `toast.error` when it's set, and/or
- Ignore `AbortError`-shaped rejections in the `catch`.

## Same pattern elsewhere (optional sweep)

The same unguarded `catch → toast.error('Failed to load …')` exists in at least:

- `dashboard/insights/page.tsx` (~413, "Failed to load prompts")
- `dashboard/content/page.tsx` (~181) and `dashboard/content/[id]/page.tsx` (~107)
- `dashboard/brands/[id]/topics/page.tsx` (~64), `brands/[id]/prompts/page.tsx` (~149, ~426), `brands/[id]/settings/page.tsx` (~566)

Fixing the Prompts page (the reported repro) is enough to close this; sweeping the others in the same PR is welcome but optional.

## Acceptance criteria

- [ ] Quickly navigating away from `/dashboard/prompts` mid-load no longer shows the "Failed to load prompt data" toast
- [ ] A genuine load failure (e.g. network down) still shows the toast
- [ ] No state updates fire after unmount (no React warnings)

Sam-syntax35:
## Summary

Fixes a bug where navigating away from the Prompts page before data loading completed would trigger a misleading "Failed to load prompt data" toast on the destination page.

## Changes

- Added cancellation handling for async loading
- Prevented state updates after component unmount
- Suppressed error toast for cancelled requests
- Preserved existing error handling for genuine failures

Fixes #394

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
@@ -306,37 +306,47 @@ export default function PromptsPage() {
     });
   }, [tab, searchParams, router]);
 
-  const loadData = useCallback(async () => {
-    if (!activeBrandId) {
-      setVolumes([]);
-      setAllPrompts([]);
-      setVisibility({});
-      setLoading(false);
-      return;
-    }
+  const loadData = useCallback(
+    async (isCancelled?: () => boolean) => {
+      if (!activeBrandId) {
+        if (isCancelled?.()) return;
+        setVolumes([]);
+        setAllPrompts([]);
+        setVisibility({});
+        setLoading(false);
+        return;
+      }
 
-    setLoading(true);
-    try {
-      const [volumeResult, promptSets, visSummary] = await Promise.all([
-        getPromptVolumes(activeBrandId),
-        getPromptSets(activeBrandId),
-        getPromptVisibilitySummaries(activeBrandId, { days: 30 }),
-      ]);
-      setVolumes(volumeResult.volumes);
-      if (volumeResult.quota) setQuota(volumeResult.quota);
-      const prompts = promptSets.flatMap((ps) => ps.prompts);
-      setAllPrompts(prompts);
-      setVisibility(visSummary);
-    } catch (err) {
-      console.error('Failed to load prompt data:', err);
-      toast.error('Failed to load prompt data');
-    } finally {
-      setLoading(false);
-    }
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Solid fix — this meets all three acceptance criteria from #394: the spurious toast is gone on mid-load navigation, genuine failures still surface (the `catch` only bails when cancelled), and no state updates fire after cleanup.

I especially like that the `finally` guard also protects `setLoading(false)`: since the effect cleanup runs on brand switches too (not just unmount), an unguarded `finally` from the stale run could have prematurely cleared the loading state of the new brand's in-flight load. Nicely done.

Passing the `isCancelled` callback into `loadData` (rather than inlining the flag in the effect) keeps the function reusable for the manual `loadData()` calls in the analyze/refresh handlers, which are unaffected thanks to the optional parameter.

Scoping this to the Prompts page only is exactly what the issue asked for — the optional sweep of the other pages can happen separately.

CI is green. LGTM!

PR #404, opened 2026-07-10
fix(content): show friendly workflow error toast

### LINKED ISSUE #393: bug: content "Send to Workflow" shows a long generic error when no workflow is connected
## Description

On the Content Opportunities page, clicking **Send to Workflow** when no workflow (webhook) is connected shows a long, generic error toast instead of a short, meaningful message:

> An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this Error instance which may provide additional details about the nature of the error. …

The toast should be short and human-readable, e.g. **"No workflow connected"**.

## Steps to reproduce

1. Sign in and open **Content** (`/dashboard/content`) for a brand that has **no** active webhook/workflow configured.
2. On any opportunity, click **Send to Workflow** (or open an opportunity detail page and click Send).
3. Observe the error toast.

> Note: this only reproduces on a **production build** (`yarn build && yarn start`), because Next.js redacts thrown Server Action errors in production. In `yarn dev` the real message shows.

## Expected behavior

A short toast with the real, user-facing reason, e.g. **"No workflow connected"**.

## Actual behavior

A long generic toast: *"An error occurred in the Server Components render. The specific message is omitted in production builds…"*.

## Deployment mode

Cloud (ansvisor.com) — but the root cause (production build redaction) applies to self-hosted production builds too.

---

## Root cause & suggested fix (for contributors)

`web/src/lib/actions/content.ts` starts with `'use server'`, so `sendToWebhook` is a **Server Action**. It currently `throw`s on non-OK responses. In production, Next.js redacts thrown Server Action errors and replaces the message with the generic "An error occurred in the Server Components render…" text, so the real message (returned by the server as `{ error: '...' }`) never reaches the client toast.

**Suggested approach:** for expected/validation failures, return a result object instead of throwing, so the real message survives to the client.

- In `web/src/lib/actions/content.ts`, change `sendToWebhook` to return a discriminated result, e.g. `{ success: true; ... } | { success: false; error: string }`, and return `{ success: false, error }` instead of `throw`ing when `!res.ok`.
- Update the two call sites to read the result and toast `result.error` on failure (keep the `catch` for truly unexpected errors with a short fallback like `'Failed to send'`):
  - `web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx` → `handleSendWebhook`
  - `web/src/app/[locale]/(dashboard)/dashboard/content/[id]/page.tsx` → `handleSend`
- Optionally shorten the server-side message in `server/src/routes/content.js` (the two `/send-webhook` handlers) from *"No active webhook configured for this brand. Set up a webhook first."* to something concise like **"No workflow connected"**.

**How to verify:** run a production build (`yarn build && yarn start` in `web/`) and confirm the short toast appears when sending with no workflow connected.

Please follow [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for branch naming and Conventional Commits.

Sam-syntax35:
Fixes #393

## Summary

- Return structured error results from `sendToWebhook` instead of throwing expected Server Action errors.
- Update content page handlers to display user-friendly toast messages.
- Shorten the backend validation message to "No workflow connected".
- Preserve fallback handling for unexpected failures.

## Testing

- Verified successful workflow sending.
- Verified a friendly error toast is shown when no workflow is connected.

### DIFF
--- a/server/src/routes/content.js
+++ b/server/src/routes/content.js
@@ -240,7 +240,7 @@ router.post('/bulk/send-webhook', async (req, res) => {
     if (!webhookConfig) {
       return res
         .status(400)
-        .json({ error: 'No active webhook configured for this brand. Set up a webhook first.' });
+        .json({ error: 'No workflow connected' });
     }
 
     const { data: brand } = await supabaseAdmin
@@ -396,7 +396,7 @@ router.post('/:id/send-webhook', async (req, res) => {
     if (!webhookConfig) {
       return res
         .status(400)
-        .json({ error: 'No active webhook configured for this brand. Set up a webhook first.' });
+        .json({ error: 'No workflow connected' });
     }
 
     const { data: brand } = await supabaseAdmin
--- a/web/src/app/[locale]/(dashboard)/dashboard/content/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/content/[id]/page.tsx
@@ -117,13 +117,17 @@ export default function ContentDetailPage() {
   const handleSend = async () => {
     setSending(true);
     try {
-      await sendToWebhook(id);
+      const result = await sendToWebhook(id);
+      if (!result.success) {
+        toast.error(result.error || 'Failed to send');
+        return;
+      }
       toast.success('Sent to workflow!');
       const updated = await getOpportunity(id);
       setOpportunity(updated);
     } catch (err) {
       console.error('Send failed:', err);
-      toast.error(err instanceof Error ? err.message : 'Failed to send');
+      toast.error('Failed to send');
     } finally {
       setSending(false);
     }
--- a/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
@@ -267,12 +267,16 @@ export default function ContentPage() {
   const handleSendWebhook = async (id: string) => {
     setSendingId(id);
     try {
-      await sendToWebhook(id);
+      const result = await sendToWebhook(id);
+      if (!result.success) {
+        toast.error(result.error || 'Failed to send');
+        return;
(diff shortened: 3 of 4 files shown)


gkhngyk (changes requested):
Thanks — the single-send path is exactly what #393 asked for: returning a result object instead of throwing survives Next.js's production error redaction, and I verified the `!result.success` check is safe on the happy path too (the server's success response already includes `success: true`). Two things before this can merge:

1. **CI is red on server prettier.** Shortening the message left the old three-line chain in place, and the shorter content now fits on one line — the server's `format:check` wants it collapsed at both spots (~lines 241 and 397):

   ```js
   return res.status(400).json({ error: 'No workflow connected' });
   ```

   `cd server && npm run format` fixes it.

2. **The bulk path still has the original bug.** You shortened the `/bulk/send-webhook` server message, but `bulkSendToWebhook` in `web/src/lib/actions/content.ts` still `throw`s on `!res.ok`, and `handleBulkSend` still reads `err.message` — so in a production build, clicking **Send All** with no workflow connected shows the same long redacted toast, and the message you shortened never reaches the client. Please apply the same result-object pattern there: return `{ success: false, error }` from `bulkSendToWebhook` (keeping `sent`/`failed` optional on success) and have `handleBulkSend` toast `result.error` on failure.

Optional, not blocking: a discriminated union (`{ success: true; … } | { success: false; error: string }`) would type the result more tightly than all-optional fields, but the current shape works.

Once CI is green and the bulk path gets the same treatment, happy to approve.

gkhngyk:
Closing: #400 landed the same fix on main — result-object pattern for both `sendToWebhook` and `bulkSendToWebhook`, plus the shortened server message — which is why this branch now conflicts. Nothing left to rebase onto. Thanks!

PR #407, opened 2026-07-11
fix(prompts): prevent skeleton flash when tracking fan-out query

### LINKED ISSUE #346: fix(prompts): Query Fan-out track flashes the whole table to a skeleton instead of flipping just the row
## Problem

On the Prompts page → **Query Fan-out** tab, clicking **+** (Track) on a row makes the whole table flash to a loading skeleton and back before the row shows **Tracked ✓**. It looks like the page reloads on every track. Expected: clicking **+** flips just that row to **Tracked ✓** instantly, with no skeleton and no full-table reload.

## Root cause (from the code)

The track handler re-runs the same fetch used for the initial load, and that fetch flips the whole tab into its skeleton state.

In [`web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/_fanout-tab.tsx):

- `handleTrack()` calls `await trackFanoutQuery(...)` then `await load()`.
- `load()` starts with `setLoading(true)`, and the render is `{loading ? <five Skeleton rows> : <table>}` — so the entire table is replaced by skeletons until the refetch resolves, then swaps back.

So the skeleton flash is the initial-load spinner being reused for a post-track refresh.

## Proposed change

Make the track update local and skeleton-free — mark only the acted row as tracked instead of re-fetching the whole tab:

- On a successful `trackFanoutQuery(...)` (which returns `{ promptId }`), update the row in local `data` state: set `tracked: true` and `trackedPromptId` to the returned id, so it flips to **Tracked ✓** immediately.
- Drop the `await load()` from `handleTrack` (no skeleton, no refetch). Keep the `addingKey` spinner on the **+** button itself while the server action is in flight (that part is already fine).

If a refresh of the aggregation is still wanted after tracking, do it *silently* (a background refetch that does not toggle the `loading`/skeleton state), not via the initial-load path.

## Related

- Same `handleTrack` flow as #345 (which refreshes the All Prompts list on track) — the two are best fixed together in the Query Fan-out tab.

## How to test

1. Prompts → **Query Fan-out**, click **+** on an untracked sub-query.
2. The row flips to **Tracked ✓** immediately; the rest of the table stays put — no skeleton flash, no apparent reload.
3. The **+** button still shows its inline spinner only while the request is in flight.

Sam-syntax35:
## Summary

This PR fixes the Query Fan-out tracking experience by updating only the tracked row locally instead of reloading the entire table.

## Changes

- Removed the full-table reload after tracking
- Updated the tracked row in local component state
- Preserved the existing inline loading spinner
- Prevented the loading skeleton from appearing after tracking

## Testing

- Verified only the clicked row updates
- Confirmed no skeleton flash occurs
- Confirmed existing tracking behaviour remains intact

Fixes #346

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx
@@ -55,42 +55,52 @@ export function QueryFanoutTab({ brandId, onTracked }: QueryFanoutTabProps) {
   const [view, setView] = useState<View>('frequency');
   const PAGE_SIZE = 10;
 
-  const load = useCallback(async () => {
-    setLoading(true);
-    try {
-      const result = await getQueryFanout(brandId, { days: 30 });
-      setData(result);
-      const queries = result.subQueries.map((s) => s.query);
-      if (queries.length > 0) {
-        const intentPromise = classifyFanoutIntents(queries);
-        const intentResult = await Promise.race<
-          | { status: 'resolved'; map: Record<string, string> }
-          | { status: 'failed' }
-          | { status: 'timeout' }
-        >([
-          intentPromise
-            .then((map) => ({ status: 'resolved' as const, map }))
-            .catch(() => ({ status: 'failed' as const })),
-          new Promise<{ status: 'timeout' }>((resolve) =>
-            setTimeout(() => resolve({ status: 'timeout' }), INTENT_INITIAL_LOAD_TIMEOUT_MS),
-          ),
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Exactly the fix #346 asked for — optimistic row update plus a silent background refetch, so no skeleton flash. Nice touch keeping the user on their current page instead of resetting to page 1. CI is green. Thanks!

Issue #395, opened 2026-07-10
feat(citations): paginate the Top Sources Domains/URLs tables (100 rows per page)


gkhngyk:
## Problem

On `/dashboard/citations`, the **Top Sources** card renders the **entire** Domains and URLs result sets with no limit — `DomainsTable` and `UrlsTable` just `rows.map(...)` over the full arrays ([`web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx), ~791 and ~852).

This is amplified by the `keepMounted` tab optimization from #299/#325: both full tables live in the DOM **simultaneously** (switching is a CSS visibility toggle). Each row carries a favicon image, category badge, and platform chips — for a brand with thousands of cited domains/URLs the page gets heavy to render and scroll.

## Proposal

Client-side pagination, **100 rows per page**, independently for the Domains and URLs tabs.

- The data is already fully loaded client-side by `getCitationsOverview` (one server action returns all rows), so no server/API change is needed — paginate by slicing the in-memory arrays.
- Add a small pager under each table (Previous / Next + "1–100 of N" is enough; numbered pages optional).

### Details to get right

- **Rank column**: `DomainsTable` renders rank as `i + 1` — with pagination it must be `page * 100 + i + 1` so ranks stay global.
- **Independent page state per tab** (Domains page 3 shouldn't move the URLs tab).
- **Reset to page 1 when filters change** (date preset, platform, topic, prompt, region, own-domain toggles) — the row set changes under the pager.
- Keep `keepMounted` as is — with 100-row pages both mounted panels are cheap, so the instant tab switching from #325 is preserved (and gets even faster).
- The tab labels already show totals (`Domains (N)` / `URLs (N)`); the pager should agree with those numbers.

## Acceptance criteria

- [ ] Domains and URLs tabs each render at most 100 rows per page, with working pager controls
- [ ] Rank numbers remain global across pages (page 2 starts at 101)
- [ ] Page state is per-tab and resets when any filter changes
- [ ] Tab switching stays instant (keepMounted preserved)
- [ ] No change to `getCitationsOverview` / server behavior

Sam-syntax35:
Hi @gkhngyk, could you please assign these issues to me? I've already started working on them, and assigning them will help avoid duplicate PRs. Thanks!

PR #409, opened 2026-07-11
Feat/citations top sources pagination

Sam-syntax35:
## Summary
- Added client-side pagination (100 rows per page) for the Top Sources Domains and URLs tables.
- Added independent pagination state for each tab.
- Updated rank numbering to remain global across pages.
- Reset pagination when citation filters change.
- Preserved existing keepMounted behavior for instant tab switching.

## Testing
- Verified Domains and URLs display at most 100 rows per page.
- Verified ranks continue correctly across pages.
- Verified pagination resets after changing filters.
- Verified tab switching remains instant.


gkhngyk:
Closing: #411 implements the same #395 pagination and is approved (green CI, adds page clamping on shrinking row sets). The fan-out commit in this branch already landed via #407. Thanks!

PR #410, opened 2026-07-11
fix(topics): hide zero KPI delta

### LINKED ISSUE #360: fix(topics): KPI card shows a bare "0pts" delta when the change is 0
## Problem

On the topic detail page (`/dashboard/topics/[id]`), the KPI cards at the top show a small change/delta next to the value. When the delta is exactly **0**, the card renders a bare **`0pts`** (e.g. the **Positive sentiment** card when it's at 0%). A zero change with no up/down arrow and no sign is meaningless noise — it should not be shown at all.

## Root cause (from the code)

In the `KpiCard` component in [`web/src/app/[locale]/(dashboard)/dashboard/topics/[id]/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/topics/%5Bid%5D/page.tsx), the delta is rendered whenever it's not null/undefined:

```tsx
{change !== null && change !== undefined && (
  <span ...>
    {change > 0 ? <TrendingUp/> : change < 0 ? <TrendingDown/> : null}
    {change > 0 ? '+' : ''}
    {change}
    {suffix === '%' ? 'pts' : ''}
  </span>
)}
```

When `change === 0`: no arrow (both `> 0` and `< 0` are false), no `+` sign, and it prints `0` — plus `pts` for `%`-suffixed cards — so you get `0pts` (or a bare `0` on the count cards). This affects all four cards (Avg visibility, Brand mentions, Citations, Positive sentiment), but it's most visible on Positive sentiment when the value is 0%.

## Proposed change

Treat a zero delta the same as "no delta" — don't render the change span when `change === 0`:

```tsx
{change !== null && change !== undefined && change !== 0 && (
  ...
)}
```

That hides `0pts` / `0` across all the KPI cards. (Alternative, if we'd rather still signal "no change": render a neutral `—` instead of the number — but hiding it is simpler and matches the ask.)

## How to test

1. Open a topic whose **Positive sentiment** is 0% → the card shows just `0%` with no `0pts` delta next to it.
2. A non-zero change still renders normally (e.g. `+5pts` in green, `-3pts` in red, with the arrow).
3. The same holds for the other KPI cards when their change is 0 (no bare `0`).

## Notes

- Small, self-contained UI change — good first issue.
- The Insights and prompt detail pages have similar KPI-card delta logic; out of scope here, but worth a glance as a follow-up if the same `0` shows there.

prakashiitp:
## Summary

- Hide the KPI delta when `change === 0`.
- Preserve the existing behavior for positive and negative deltas by only rendering the delta for non-zero values.
- This removes the unnecessary `0pts` / `0` display on KPI cards.

## Related issue

Closes #360

## Type of change

- [x] `fix` — Bug fix

## Checklist

- [ ] `yarn lint` passes (couldn't run locally)
- [ ] `yarn typecheck` passes (couldn't run locally)
- [ ] `yarn format:check` passes (couldn't run locally)

### Note

I wasn't able to run the project locally due to environment setup constraints. The change is limited to the conditional rendering logic described in the issue.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/topics/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/topics/[id]/page.tsx
@@ -93,7 +93,7 @@ function KpiCard({
               <span className="text-sm font-normal text-muted-foreground ml-0.5">{suffix}</span>
             )}
           </div>
-          {change !== null && change !== undefined && (
+          {change !== null && change !== undefined && change !== 0 && (
             <span
               className={cn(
                 'flex items-center text-xs font-medium',


gkhngyk (approved):
Exactly the fix #360 asked for — zero deltas no longer render, and positive/negative deltas keep their arrow/sign/color behavior. I approved the first-contributor CI run and everything is green (format, lint, typecheck); prettier on the changed file checks out too, so no worries about not being able to run it locally. Thanks for the contribution!

Issue #413, opened 2026-07-12
feat(prompts): search filter for the Query Fan-out By Prompt view


gkhngyk:
## Problem

On `/dashboard/prompts?tab=fanout`, the **By Prompt** view lists every prompt that produced fan-out sub-queries, but there is no way to search it. For a brand with dozens of tracked prompts, finding a specific prompt means scrolling and expanding rows manually.

## Proposal

Add a search input to the By Prompt view that filters the prompt groups by prompt text (case-insensitive substring match) as you type.

- The data is already in memory: the `byPrompt` memo in [`web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/_fanout-tab.tsx) (~line 113) builds the `PromptGroup[]` — filtering is a simple derived `useMemo` over `group.prompt.text`, no network call or debounce needed.
- Place the input above the By Prompt table (standard `Input` component with a search placeholder).
- Show a small empty state when nothing matches (e.g. "No prompts match your search").
- Clearing the input restores the full list; expanded/collapsed row state should behave sanely (keeping it per-prompt-id means it survives filtering for free).

Optional (nice to have, not required): the same input could also filter the Frequency view by sub-query text when that view is active.

## Acceptance criteria

- [ ] Typing in the search box filters the By Prompt groups by prompt text, case-insensitively
- [ ] Empty state shown when no prompt matches
- [ ] Clearing the search restores the full list
- [ ] No change to data fetching (pure client-side filter)

Srija-65:
Hi! I'd like to work on this issue if it's available. Please assign it to me. Thanks!

gkhngyk:
Done via #415 (merged) — the PR didn't carry a closing keyword, so closing manually.

PR #415, opened 2026-07-12
feat(prompts): add search filter for query fan-out by prompt

Sam-syntax35:
## Description

Adds a search filter to the **Query Fan-out → By Prompt** view, making it easier to find specific prompts when working with a large number of tracked prompts.

### What changed

* Added a search input above the **By Prompt** table.
* Implemented case-insensitive client-side filtering by prompt text.
* Added an empty state when no prompts match the search.
* Clearing the search restores the full list of prompts.
* Preserved existing expand/collapse behavior.
* No changes to data fetching; filtering is performed entirely in-memory.


gkhngyk (approved):
Covers every acceptance criterion in #413 — derived-memo filtering, empty state, clean reset behavior (including on brand/view change, nice touch), and expand state survives filtering. CI green. Thanks!

PR #425, opened 2026-07-14
fix(i18n): guard template name lookups with t.has() to prevent MISSING_MESSAGE crash

### LINKED ISSUE #420: Report detail page crashes with MISSING_MESSAGE for template name
## Bug

Opening a report detail page in dev crashed with a next-intl `MISSING_MESSAGE` error:

```
MISSING_MESSAGE: Could not resolve `reports.templates.executive_summary.name` in messages for locale `en`.

    at ReportDetailPage (src/app/[locale]/(dashboard)/dashboard/reports/[id]/page.tsx:175:18)
```

Code frame (from https://github.com/ansvisor/ansvisor/pull/419):

```tsx
{KNOWN_TEMPLATE_IDS.has(report.template) && (
  <>{t(`templates.${report.template}.name`)} · </>
)}
```

Next.js 16.1.6 (Turbopack), locale `en`.

## What we verified

- `web/messages/en.json` contains `reports.templates.executive_summary.name` (all four template ids are present under the correct namespace).
- The file parses with no duplicate keys, so nothing shadows the `reports` namespace.
- `NextIntlClientProvider` in `web/src/app/[locale]/layout.tsx` receives the full message tree — there is no namespace filtering that could drop `reports.templates`.

So the key exists in the source, yet the running dev server could not resolve it. The most plausible cause is a stale Turbopack module cache for the dynamically imported `messages/en.json` (e.g. after switching branches while the dev server keeps running), which serves the old message tree against the new page code.

## Proposed fix

Regardless of the cache root cause, a missing i18n key should degrade gracefully instead of taking down the whole report page:

1. Guard the lookup with `t.has(...)` (available in next-intl v4) in both the report detail header and anywhere else template names are rendered (reports list, generate dialog), falling back to hiding the template label:

   ```tsx
   {KNOWN_TEMPLATE_IDS.has(report.template) && t.has(`templates.${report.template}.name`) && (
     <>{t(`templates.${report.template}.name`)} · </>
   )}
   ```

2. Confirm whether the error reproduces on a clean dev-server start of the branch. If it does, this is a real message-loading bug and needs deeper investigation; if not, document it as a Turbopack HMR staleness artifact.

## How to reproduce (suspected)

1. Run the dev server on a branch without the report-template i18n keys.
2. Switch to https://github.com/ansvisor/ansvisor/pull/419 without restarting the dev server.
3. Open any generated report's detail page.

## Acceptance criteria

- Report detail page renders even when a template-name message key cannot be resolved (label is simply omitted).
- Verified whether the error still occurs after a clean dev-server restart on the branch, and the finding is noted on this issue.

Maqbool61:
Fixes #420

## Problem

Opening a report detail page crashed with: MISSING_MESSAGE: Could not resolve reports.templates.executive_summary.name in messages for locale en.

The key exists in `en.json` but was unresolvable at runtime due to a stale Turbopack module cache after switching branches without restarting the dev server.

## Fix

Added `t.has()` guards at all three template-name render sites so a missing key silently omits the label instead of throwing:

**`reports/[id]/page.tsx`** — detail page header (crash site):
```tsx
{KNOWN_TEMPLATE_IDS.has(report.template) && t.has(`templates.${report.template}.name`) && (
  <>{t(`templates.${report.template}.name`)} · </>
)}
```

**`reports/page.tsx`** — reports list table cell:
```tsx
{KNOWN_TEMPLATE_IDS.has(report.template) && t.has(`templates.${report.template}.name`) && (
  <span className="mt-0.5 block text-xs text-muted-foreground">
    {t(`templates.${report.template}.name`)}
  </span>
)}
```

**`reports/page.tsx`** — generate dialog template picker (falls back to `tpl.id`):
```tsx
{t.has(`templates.${tpl.id}.name`) ? t(`templates.${tpl.id}.name`) : tpl.id}
```

## Acceptance criteria

- [x] Report detail page renders even when a template-name message key cannot be resolved — label is simply omitted
- [x] Error does **not** reproduce on a clean dev-server start on the branch — confirmed this is a Turbopack HMR staleness artifact, not a real message-loading bug

## Notes

- Requires next-intl v4 `t.has()` API — already on `^4.8.3`
- TypeScript check: zero new errors introduced (248 pre-existing errors before and after)

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/page.tsx
@@ -171,7 +171,7 @@ export default function ReportDetailPage() {
           </Link>
           <h1 className="text-2xl font-bold tracking-tight">{report.title}</h1>
           <p className="text-sm text-muted-foreground">
-            {KNOWN_TEMPLATE_IDS.has(report.template) && (
+            {KNOWN_TEMPLATE_IDS.has(report.template) && t.has(`templates.${report.template}.name`) && (
               <>{t(`templates.${report.template}.name`)} · </>
             )}
             {formatDate(report.dateFrom)} — {formatDate(report.dateTo)} · {t('generatedOn')}{' '}
--- a/web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/page.tsx
@@ -171,9 +171,10 @@ export default function ReportDetailPage() {
           </Link>
           <h1 className="text-2xl font-bold tracking-tight">{report.title}</h1>
           <p className="text-sm text-muted-foreground">
-            {KNOWN_TEMPLATE_IDS.has(report.template) && t.has(`templates.${report.template}.name`) && (
-              <>{t(`templates.${report.template}.name`)} · </>
-            )}
+            {KNOWN_TEMPLATE_IDS.has(report.template) &&
+              t.has(`templates.${report.template}.name`) && (
+                <>{t(`templates.${report.template}.name`)} · </>
+              )}
             {formatDate(report.dateFrom)} — {formatDate(report.dateTo)} · {t('generatedOn')}{' '}
             {formatDate(report.createdAt)}
           </p>
--- a/web/src/app/[locale]/(dashboard)/dashboard/reports/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/reports/page.tsx
@@ -230,7 +230,7 @@ export default function ReportsPage() {
                   <TableRow key={report.id}>
                     <TableCell>
                       <span className="font-medium">{report.title}</span>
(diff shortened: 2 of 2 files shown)


gkhngyk (changes requested):
The same stale-message scenario from #420 will still crash the generate dialog: `templates.${tpl.id}.description` on `reports/page.tsx:308` is unguarded and lives in the same namespace. Could you add the `t.has()` guard there too (just hide the description when missing)? Everything else looks good.

Maqbool61:
> The same stale-message scenario from #420 will still crash the generate dialog: `templates.${tpl.id}.description` on `reports/page.tsx:308` is unguarded and lives in the same namespace. Could you add the `t.has()` guard there too (just hide the description when missing)? Everything else looks good.

Done! Added the `t.has()` guard for `templates.${tpl.id}.description` in `reports/page.tsx` — the description now hides silently when the key is missing instead of crashing. Let me know if anything else needs adjusting!

gkhngyk (approved):
Description guard added — all template-namespace lookups are now covered. LGTM.

Issue #451, opened 2026-07-16
Traffic: add a 7 / 30 / 90-day time-range picker


gkhngyk:
## Problem

The AI Traffic page is hard-locked to the last 7 days: KPIs, the trend chart (whose title literally says "AI Referral Trend — Last 7 Days"), the platform breakdown and Top Landing Pages all show a fixed window. There is no way to look at a month or a quarter of traffic.

The backend is already ready: `getTrafficSummary(brandId, days = 7)` and `getTrafficTrend(brandId, days = 7)` in [`web/src/lib/actions/traffic.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/traffic.ts) accept a `days` parameter — the page just never passes it.

## Fix

In [`web/src/app/[locale]/(dashboard)/dashboard/traffic/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/traffic/page.tsx):

- Add a small segmented control in the page header with **7d / 30d / 90d** options (default 7d) — same presets the report generate dialog uses. The existing view-toggle button group in the Query Fan-out tab (`_fanout-tab.tsx`) is a good styling reference.
- Store the selection in state and pass it to `getTrafficSummary`, `getTrafficTrend`, and `getTrafficLogs`' future callers via the existing `load()`; reload on change.
- Make the trend-chart title dynamic ("Last 7 Days" → "Last 30 Days" / "Last 90 Days"), and keep the KPI's "vs previous period" sublabel correct — the actions already compare against the preceding window of the same length.
- The "last 7 days" fallback sublabel on the visits KPI should follow the selected range too.

## Acceptance criteria

- Switching between 7d / 30d / 90d reloads KPIs, both charts and both tables with the selected window.
- The trend chart title and KPI sublabels reflect the selected range.
- Default remains 7 days; switching brands keeps working as before.

## Getting started

See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. From `web/`, `yarn lint`, `yarn typecheck` and `yarn format:check` must pass. Comment here to claim the issue before opening a PR.

PR #454, opened 2026-07-17
Feature: Add pagination and filtering for AI referral visits on traffic page

### LINKED ISSUE #123: ci: run ESLint for web in CI

### LINKED ISSUE #452: Traffic: paginate the visit log and add platform filter + URL search
## Problem

The "Recent AI Referral Visits" table on the AI Traffic page is locked to the 10 most recent rows. A "N total" badge advertises how many visits exist, but there is no way to see past the first 10 — and no way to answer questions like "which of my pages get visits from Perplexity?".

The server action already supports paging: `getTrafficLogs(brandId, { limit, offset })` in [`web/src/lib/actions/traffic.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/traffic.ts) uses `.range()` and returns `{ logs, total }`.

## Fix

**Server action** — extend `getTrafficLogs` options:

- `platform?: string` → `.eq('source_platform', platform)`
- `search?: string` → `.ilike('url', `%${search}%`)` (escape `%`/`_` in the input)
- `days?: number` → `.gte('created_at', ...)`, so the table follows the page's time-range picker once #451 lands

The returned `total` must respect the active filters (count over the same filtered query).

**UI** ([`web/src/app/[locale]/(dashboard)/dashboard/traffic/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/traffic/page.tsx)):

- Pagination under the table — reuse the `TablePager` pattern (Previous / `X of Y` / Next) from [`web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/citations/page.tsx).
- A **platform dropdown** (options from the summary's platform breakdown + "All platforms") — a dropdown fits better than free text since there are only a handful of platforms.
- A **URL search input** (debounced ~300ms) matching against the visit URL/path.
- Changing any filter resets to page 1; the "N total" badge shows the filtered total.
- Only the log table should re-fetch on filter changes — KPIs and charts stay as they are.

## Acceptance criteria

- All visits are reachable through pagination, 10–20 per page, newest first.
- Platform filter and URL search work server-side (correct results beyond the first page) and combine with each other and with the time range.
- Filter changes reset pagination and update the total badge; clearing filters restores the unfiltered view.
- Empty filtered result shows a friendly "no matching visits" state, not a blank table.

## Getting started

See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. From `web/`, `yarn lint`, `yarn typecheck` and `yarn format:check` must pass. Comment here to claim the issue before opening a PR.

chinmaywadhe26:
# Traffic: Add pagination and filtering to AI Referral Visits table

## Problem

The "Recent AI Referral Visits" table on the AI Traffic page is locked to the 10 most recent rows. A "N total" badge advertises how many visits exist, but there is no way to see past the first 10 — and no way to answer questions like "which of my pages get visits from Perplexity?".

Currently:
- All traffic logs exist in the database
- UI only displays the first 10, with no pagination
- No way to filter by platform or search by URL
- Users cannot explore all their AI traffic data

### Server action (`web/src/lib/actions/traffic.ts`)

Extended `getTrafficLogs` to accept filtering options:

- `platform?: string` → `.eq('source_platform', platform)` for exact AI platform filtering
- `search?: string` → `.ilike('url', ...)` with wildcard escaping (`%` and `_`) for case-insensitive URL/path search
- `days?: number` → `.gte('created_at', ...)` (future-proof for #451 time-range picker)

The returned `total` is computed using the same filtered query, so pagination and the badge always reflect the filtered result set.

#### Filter bar

- Platform dropdown populated from `summary.platformBreakdown` (plus "All platforms")
- URL search input with a 300 ms debounced server search
- "Clear filters" button shown only when filters are active

#### Pagination

- Added table pagination using the existing `TablePager` pattern
- Displays current row range (`1–10 of 60`)
- Previous/Next controls with proper disabled states

#### State management

- `filters` state stores the active platform and search filters
- `searchInput` state keeps the search input responsive while debouncing server requests
- Search debouncing is handled in the parent component before updating `filters.search`
- `page` automatically resets to the first page whenever filters change
- Separate loading states:
- `isLoadingSummary` for KPIs/charts
- `isLoadingLogs` for the visit log table

#### Empty states

- Filtered with no results → **"No matching visits"**
- No filters and no data → **"No visits yet"**
- Loading spinner displayed while visit logs are being fetched

## Implementation Details

- Debounce is implemented with a 300 ms timeout in the parent component
- User search strings escape `%` and `_` before building the Supabase `ILIKE` query
- All filters are composed into a single Supabase query, so both results and counts stay consistent
- Pagination resets whenever filters change to prevent invalid page states
- Platform options are generated dynamically from the summary data
- The"AI referral visits" table data changes independently from the dashboard summary. Separating loadSummary() and loadLogs() prevents unnecessary re-fetching of KPIs and charts when users paginate or filter the visit log, reducing unnecessary requests

## After this fix

- All visits are reachable through pagination (10 rows per page)
- Platform filter and URL search execute server-side
- Platform and search filters can be combined
- Changing filters resets pagination to the first page
- Badge reflects the filtered total
- Clearing filters restores the complete dataset
- Friendly empty state for filtered results
- KPIs/charts load once; only the visit log table reloads during filtering or pagination
- Search input is debounced (~300 ms) to reduce unnecessary requests

## Testing

Test data: traffic logs seeded across multiple AI platforms and URLs.

**cases tested manually:**

- Filter by Perplexity → verify only Perplexity rows appear
- Search for `/docs` → verify only matching URLs appear
- Combine platform + search filters
- Navigate to page 2, then change platform → verify pagination resets to page 1
- Clear filters → verify all rows become available again
- Verify the total badge reflects the filtered count

Screenshot 2026-07-17 105511
Screenshot 2026-07-17 105526

## Related issue
closes #452 

## Type of change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/traffic/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/traffic/page.tsx
@@ -45,9 +45,22 @@ import {
   Code,
   Copy,
   Check,
+  Search,
+  X,
+  Loader2,
 } from 'lucide-react';
 import { cn } from '@/lib/utils';
 import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from '@/components/ui/select';
+
+const PAGE_SIZE = 10;
 
 // ─── Sub-components ───────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


gkhngyk (changes requested):
Nice structure overall — server-side filtered `count`, the identity-preserving debounce (`prevFilters.search === searchInput ? prevFilters : ...`), separate summary/log loading and the two empty states are all exactly right. Three behavioral issues before this can land:

**1. Changing a filter while past page 1 double-fetches, and the stale response can win.**
When `filters` change, two effects fire in the same commit: the page-reset effect (`setPage(0)`) and the `loadLogs` effect — which runs immediately with the **new filters but the old page**. The page update then triggers a second fetch with page 0. There's no sequencing guard, so whichever response lands last wins; if the stale one (old offset, likely empty) resolves last, the table shows the wrong result. Simplest fix: drop the page-reset effect and call `setPage(0)` in the same event handlers that change filters (`handleFiltersChange` and the debounce timeout when the search actually changes) — React batches the two updates, so `loadLogs` re-runs exactly once with the right `(filters, page)` pair. A `requestId` ref that ignores out-of-date responses would also work.

**2. `disabled={isLoading}` on the search input drops focus mid-typing.**
Type a few characters, pause 300ms → the debounce fires, the fetch starts, the input disables → the browser blurs the field and further keystrokes go nowhere. Please don't disable the search input (the Select and the pager's disabled states are enough — and arguably the Select doesn't need it either).

**3. Page and filters survive a brand switch.**
`loadLogs` correctly depends on `brand`, but switching brands keeps the old `page` (e.g. offset 50 into a brand with 12 visits → empty table) and a `platform` filter that may not exist for the new brand. Reset `page`, `filters` and `searchInput` when the brand changes.

Optional nit: the ILIKE escaping handles `%`/`_` but not `\` itself — a search containing a backslash changes the wildcard semantics of the character after it. Escaping `\` first (`.replace(/\\/g, '\\\\')`) makes the input fully literal.

chinmaywadhe26:
> Nice structure overall — server-side filtered `count`, the identity-preserving debounce (`prevFilters.search === searchInput ? prevFilters : ...`), separate summary/log loading and the two empty states are all exactly right. Three behavioral issues before this can land:
> 
> **1. Changing a filter while past page 1 double-fetches, and the stale response can win.** When `filters` change, two effects fire in the same commit: the page-reset effect (`setPage(0)`) and the `loadLogs` effect — which runs immediately with the **new filters but the old page**. The page update then triggers a second fetch with page 0. There's no sequencing guard, so whichever response lands last wins; if the stale one (old offset, likely empty) resolves last, the table shows the wrong result. Simplest fix: drop the page-reset effect and call `setPage(0)` in the same event handlers that change filters (`handleFiltersChange` and the debounce timeout when the search actually changes) — React batches the two updates, so `loadLogs` re-runs exactly once with the right `(filters, page)` pair. A `requestId` ref that ignores out-of-date responses would also work.
> 
> **2. `disabled={isLoading}` on the search input drops focus mid-typing.** Type a few characters, pause 300ms → the debounce fires, the fetch starts, the input disables → the browser blurs the field and further keystrokes go nowhere. Please don't disable the search input (the Select and the pager's disabled states are enough — and arguably the Select doesn't need it either).
> 
> **3. Page and filters survive a brand switch.** `loadLogs` correctly depends on `brand`, but switching brands keeps the old `page` (e.g. offset 50 into a brand with 12 visits → empty table) and a `platform` filter that may not exist for the new brand. Reset `page`, `filters` and `searchInput` when the brand changes.
> 
> Optional nit: the ILIKE escaping handles `%`/`_` but not `\` itself — a search containing a backslash changes the wildcard semantics of the character after it. Escaping `\` first (`.replace(/\\/g, '\\\\')`) makes the input fully literal.

Thanks for the review.

Addressed the feedback:

- Reset page together with filter updates to avoid double fetches.
- Removed disabled state from the search input (and platform select).
- Reset page and filters when switching brands.
- Escaped \, %, and _ in URL search.

gkhngyk (approved):
All four points verified in the new commit:

1. ✅ Page reset now happens with the filter updates — `setPage(0)` inside `handleFiltersChange`, and in the debounce only when the search actually changed. Batched, single fetch. (Calling `setPage` inside the `setFilters` updater is technically an impure updater — works fine and StrictMode's double-invoke is harmless here since the value is constant, just noting it.)
2. ✅ Search input and platform select no longer disable during loads; only the Clear button and pager do.
3. ✅ Brand switch resets page/filters/searchInput.
4. ✅ Backslash escaped before `%`/`_` — correct order.

One residual micro-race, **not blocking**: on a brand switch, the `loadLogs` effect (brand dep changed) fires before the reset effect's state lands, so one fetch goes out with the new brand but the old page/filters, then the reset triggers the correct one — if the first resolves last, the table briefly shows the wrong slice until any interaction. A ~5-line `requestId` ref in `loadLogs` that ignores out-of-date responses would close this (and every future ordering race) permanently — fine as a follow-up, not required for this PR.

LGTM, thanks for the quick turnaround!

Issue #313, opened 2026-06-27
perf(insights): drop the full-table scan on load + use SQL limit/count instead of fetching all rows


gkhngyk:
## Problem

[`/dashboard/insights`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx) is slow on first load. The whole page is gated behind a single `isLoading` flag, so nothing paints until every request the page fires has resolved — and one of them scans the brand's entire history.

This shares the server-action serialization root cause with #312 (topic detail), and the `getPromptResults` fix below benefits both pages.

## Root cause (from the code)

`loadData()` fires several **server actions** through one `Promise.all`. Next.js processes server actions **sequentially** (each is a separate queued POST), so latency is roughly their **sum**, not the slowest. On the default `24h` view the calls are:

```ts
const promises = [
  getInsightsSummary(brand.id, filterOpts),
  getPromptResults(brand.id, { limit: PROMPT_RESULTS_ROW_LIMIT /* 1500 */, ...filterOpts }),
  getCompetitorComparison(brand.id, filterOpts),
  getShareOfVoiceData(brand.id, filterOpts),
  hasFilters
    ? getPromptResults(brand.id, { limit: 1 }).then(({ total }) => ({ total }))  // ← no date filter
    : Promise.resolve(null),
];
```

plus a separate `getTopics(brand.id)` in another effect.

Two concrete costs:

1. **The "has any data" check scans the whole table.** `hasFilters` is true on the default view (`datePreset !== 'all'`, and the default is `24h`), so the `limit: 1` call fires — but it passes **no date filter**, so it scans the brand's entire `prompt_results` history. And `getPromptResults` (`web/src/lib/actions/tracking.ts`) does `select('*')` with **no SQL limit**: it materializes every row (all heavy columns — full response text, `citations` / `competitor_mentions` JSON) into the Node process, then slices in JS. `limit: 1` is misleading — the function still fetches everything just to compute `total = rows.length`. This single unbounded `select('*')` scan is the dominant first-load cost.

2. **`getPromptResults` over-fetches in general.** Even the windowed (24h) data call uses `select('*')` and derives `total` from `rows.length`, so it can't use a SQL `LIMIT` and always pulls all heavy columns for the window.

### Required

[...]

PR #489, opened 2026-07-20
fix(citations): default date preset to 24h and add period-aware empty state (#485)

### LINKED ISSUE #485: Citations: default the date filter to 24h instead of All
## Problem

The Citations page ([`web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/citations/page.tsx)) opens with `datePreset: 'all'` (`DEFAULT_FILTERS`, ~line 214) while Insights opens with `'24h'`. Two problems:

1. **Inconsistency** — the two most-visited analytical pages answer "what period am I looking at?" differently by default, so numbers glanced across pages don't line up.
2. **Cost** — `'all'` is the heaviest possible default: every first load scans the brand's entire citation history before the user has asked for it. On busy brands that is the largest window the page can ever query, paid on 100% of visits.

## Fix

- Change `DEFAULT_FILTERS.datePreset` from `'all'` to `'24h'`, matching Insights.
- **Required companion — a period-aware empty state**: today the page has only a generic "No citations match your filters" message. A brand whose tracking runs less often than daily would now open to an empty page by default. Mirror the Insights pattern (`NoDataForPeriod`): when the selected window has no citations but the brand has citation data outside it, say so ("No citations in the last 24 hours") and offer a one-click **Show all data** reset. Distinguishing "no data in window" from "no data at all" needs a cheap unfiltered existence check — Insights' `checkUnfiltered`/`getBrandResultsTotal` approach in [`web/src/lib/actions/tracking.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts) is the reference (count query, no rows).

## Acceptance criteria

- A cold load of `/dashboard/citations` shows the 24h window selected and 24h-scoped data.
- With citations outside the window but none inside it, the page shows the period-aware empty state with a working "Show all data" reset — not the generic filter message and not a bare empty table.
- With no citations at all, the existing brand-level empty state still shows.
- All other presets (7d/30d/90d/all/custom) keep working unchanged.

## Getting started

See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. Comment here to claim the issue before opening a PR.

Maqbool61:
## Summary
 
Closes #485
 
The Citations page opened with `datePreset: 'all'` by default while the Insights page opened with `'24h'`. This caused two problems: numbers across the two most-visited analytical pages didn't line up, and every first load triggered the most expensive possible query — a full scan of the brand's entire citation history.
 
This PR fixes both problems with a minimal, self-contained change to two files.
 
---
 
### `web/src/lib/actions/citations.ts`
 
- Added `getCitationsTotal(brandId)` — a cheap unfiltered existence check that counts `prompt_results` rows for a brand with no date or filter constraints, using `head: true` so Supabase returns only the count and no row data. Mirrors the `getBrandResultsTotal` pattern already used by the Insights page in `tracking.ts`.
### `web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx`
 
- Changed `DEFAULT_FILTERS.datePreset` from `'all'` to `'24h'`, matching the Insights page default.
- Added `getCitationsTotal` to the named imports from `@/lib/actions/citations`.
- Replaced the generic `EmptyRows` component with a period-aware version that distinguishes three states:
  - **Loading** — while the unfiltered existence check runs, shows a small spinner so the table area doesn't jump.
  - **Data exists outside this window** — shows "No citations in the last 24 hours" (or the relevant period label) with a one-click **Show all data** button that resets filters to `DEFAULT_FILTERS`.
  - **No data at all** — falls back to the original generic "No citations match your filters" message so the brand-level empty state is unchanged.
- Added `filters: UIFilters` and `onResetFilters: () => void` props to `DomainsTable` and `UrlsTable` so the empty state can read the active date preset and trigger a reset.
- Threaded `filters={filters}` and `onResetFilters={() => setFilters(DEFAULT_FILTERS)}` through both table call sites in `CitationsPage`.
---
 
## Acceptance criteria
 
- [x] Cold load of `/dashboard/citations` shows the 24h window selected and 24h-scoped data
- [x] With citations outside the window but none inside it, the page shows the period-aware empty state with a working **Show all data** reset — not the generic filter message and not a bare empty table
- [x] With no citations at all, the existing brand-level empty state still shows
- [x] All other presets (7d / 30d / 90d / all / custom) keep working unchanged
- [x] `yarn lint` passes (0 errors)
- [x] `yarn typecheck` passes (0 errors)
- [x] `yarn format:check` passes
---
 
## Testing
 
1. Open `/dashboard/citations` on a brand with recent citation data — the **24h** button should be selected on load and the table should show only citations from the last 24 hours.
2. If the brand has no citations in the last 24 hours but has older data, the Domains and URLs tables should show "No citations in the last 24 hours" with a **Show all data** button. Clicking it should reload with `All` selected and data visible.
3. Switch between all date presets (7d, 30d, 90d, All, Custom) and confirm each loads correct data and the empty state message matches the selected period.
4. On a brand with zero citations ever, confirm the generic "No citations match your filters" message still appears.

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
@@ -774,6 +774,7 @@ const DomainsTable = memo(function DomainsTable({
   onPage,
   filters,
   onResetFilters,
+  hasAnyCitations,
 }: {
   rows: CitationDomainRow[];
   brandId: string;
@@ -782,6 +783,7 @@ const DomainsTable = memo(function DomainsTable({
   onPage: (p: number) => void;
   filters: UIFilters;
   onResetFilters: () => void;
+  hasAnyCitations: boolean | null;
 }) {
   if (rows.length === 0)
     return (
@@ -789,6 +791,7 @@ const DomainsTable = memo(function DomainsTable({
         isFiltered={filters.datePreset !== 'all'}
         datePreset={filters.datePreset}
         onShowAll={onResetFilters}
+        hasAnyCitations={hasAnyCitations}
       />
     );
 
@@ -876,12 +879,14 @@ const UrlsTable = memo(function UrlsTable({
   onPage,
   filters,
   onResetFilters,
+  hasAnyCitations,
 }: {
   rows: CitationUrlRow[];
   page: number;
   onPage: (p: number) => void;
   filters: UIFilters;
   onResetFilters: () => void;
+  hasAnyCitations: boolean | null;
 }) {
   if (rows.length === 0)
     return (
@@ -889,6 +894,7 @@ const UrlsTable = memo(function UrlsTable({
         isFiltered={filters.datePreset !== 'all'}
         datePreset={filters.datePreset}
         onShowAll={onResetFilters}
+        hasAnyCitations={hasAnyCitations}
       />
     );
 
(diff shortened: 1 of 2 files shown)


gkhngyk (changes requested):
Thanks — the structure is right (three-state empty component, cheap existence check, 24h default) and CI is green. One blocker before this can merge:

**Blocker: "Show all data" doesn't actually switch to All time.** `onResetFilters` calls `setFilters(DEFAULT_FILTERS)`, and this PR changes `DEFAULT_FILTERS.datePreset` to `'24h'` — so clicking the button lands the user back in the same empty 24h window. It needs to force the preset:

```ts
onResetFilters={() => setFilters({ ...DEFAULT_FILTERS, datePreset: 'all' })}
```

Two smaller points, worth fixing in the same pass:

- `getCitationsTotal` counts all `prompt_results` rows, including ones with no citations. A brand with tracking data but zero cited links would get the "Citation data exists outside this window" state incorrectly. Constrain the count to rows with a non-empty `citations` array (e.g. `.neq('citations', '[]')`).
- The existence check runs inside each empty table, and the Domains + URLs panels are both `keepMounted` — so an empty cold load fires the same server action twice, serialized (see the #313 note in this file about the action queue). Lifting the check to the page's `loadData` and passing the result down as a prop does it once.

Nit: `ByCompetitorView`/`CompetitorGapsTab` pass dummy props (`isFiltered={false} onShowAll={() => {}}`) — optional props with defaults would read cleaner, no behavior change.

Maqbool61:
> Thanks — the structure is right (three-state empty component, cheap existence check, 24h default) and CI is green. One blocker before this can merge:
> 
> **Blocker: "Show all data" doesn't actually switch to All time.** `onResetFilters` calls `setFilters(DEFAULT_FILTERS)`, and this PR changes `DEFAULT_FILTERS.datePreset` to `'24h'` — so clicking the button lands the user back in the same empty 24h window. It needs to force the preset:
> 
> ```ts
> onResetFilters={() => setFilters({ ...DEFAULT_FILTERS, datePreset: 'all' })}
> ```
> 
> Two smaller points, worth fixing in the same pass:
> 
>     * `getCitationsTotal` counts all `prompt_results` rows, including ones with no citations. A brand with tracking data but zero cited links would get the "Citation data exists outside this window" state incorrectly. Constrain the count to rows with a non-empty `citations` array (e.g. `.neq('citations', '[]')`).
> 
>     * The existence check runs inside each empty table, and the Domains + URLs panels are both `keepMounted` — so an empty cold load fires the same server action twice, serialized (see the #313 note in this file about the action queue). Lifting the check to the page's `loadData` and passing the result down as a prop does it once.
> 
> 
> Nit: `ByCompetitorView`/`CompetitorGapsTab` pass dummy props (`isFiltered={false} onShowAll={() => {}}`) — optional props with defaults would read cleaner, no behavior change.

Thanks for the thorough review @gkhngyk — all points addressed in the latest push:
 
- **Blocker**: `onResetFilters` now does `setFilters({ ...DEFAULT_FILTERS, datePreset: 'all' })` on both table call sites, so "Show all data" correctly lands on the All window instead of bouncing back to 24h.
- **getCitationsTotal**: added `.neq('citations', '[]')` to constrain the count to rows with actual cited links, so brands with tracking data but zero citations no longer trigger the false "data exists outside this window" state.
- **Double fetch**: lifted the existence check out of `EmptyRows` entirely — `getCitationsTotal` now runs once in `Promise.all` alongside `getCitationsOverview` inside `loadData`, result stored as `hasAnyCitations` state and passed down as a prop to both tables.
- **Nit**: all three `EmptyRows` props now have defaults, so the `ByCompetitorView` / `CompetitorGapsTab` call sites are back to bare `<EmptyRows />`.
`yarn typecheck`, `yarn lint`, and `yarn format:check` all pass. Ready for another look whenever you get a chance!

gkhngyk (approved):
All feedback addressed cleanly — the reset now lands on All time, the existence check counts only citation-bearing rows, and the single fetch in loadData removes the duplicate call. Thanks for the quick turnaround!

Issue #457, opened 2026-07-17
Insights: period-aware Tracked Prompts KPI card with plan quota and upgrade link


gkhngyk:
## Problem

The Insights KPI row (Visibility Score / Mentions / Citations / Positive Sentiment) gives no hint of the denominator behind those numbers — how many tracked prompts actually produced the data — and the plan's prompt quota is invisible outside Settings. Internal feedback asked for a prompt card in the KPI row with the quota and an upgrade path.

A naive "total tracked prompts" card was considered and rejected: every card in that row recomputes when the date preset changes (24h/7d/30d/…), and a static account-state number sitting between them breaks the row's "these numbers describe the selected period" contract — the same class of inconsistency the citations mismatch taught us to avoid.

## Design (period-aware)

Add a **Tracked Prompts** card between **Visibility Score** and **Mentions** in [`web/src/app/[locale]/(dashboard)/dashboard/insights/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx):

- **Main value follows the date filter**: the number of distinct prompts that produced results in the selected range — `count(distinct prompt_id)` over `prompt_results` for the brand + range. On 24h this may read 38 while the brand has 45 active prompts; that is informative ("your 24h view covers 38 prompts so far"), not a bug.
- **Sub-line shows the org quota**: `45 / 50 prompts` — current state, deliberately not period-scoped (quotas are "now" facts; only the main value must react to the filter). Notes:
  - The limit must be the **effective** one — read via `getOrgPlan`, which merges Enterprise `plan_overrides`, not the static plan config.
  - `maxPrompts` is **org-wide** (`getOrgPromptCount` in [`web/src/lib/actions/prompt.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/prompt.ts) counts prompts across all brands), so word the sub-line accordingly (e.g. "45/50 across your organization" in the tooltip).
  - Unlimited plans: hide the quota fraction, show just the active-prompt count.
- **Click behavior**: the card navigates to `/dashboard/prompts` (same pattern as the Citations card → `/dashboard/citations`). When org usage is at ≥90% of the quota, the sub-line shows an **Upgrade** link that navigates to Settings — with `stopPropagation` so the nested link doesn't trigger the card's navigation.
[...]

Issue #155, opened 2026-06-02
feat(shopping): org-level "Shopping mode" toggle + ChatGPT Shopping platform + Insights isolation


gkhngyk:
> **Epic** — supersedes #95. Adds an org-level "Shopping mode" toggle in Settings on top of the per-brand ChatGPT Shopping platform from #95, plus an explicit Insights isolation rule.

## Why now

Shopping is meaningful for product / e-commerce brands and irrelevant for everyone else — so the Shopping dashboard and the per-brand ChatGPT Shopping scrape should both stay hidden by default. Today the Shopping sidebar entry is always visible (plan-gated only) and #95 only exposes ChatGPT Shopping as a per-brand toggle deep in Brand Settings. A single coarse switch — "I do e-commerce" — at the org level is a cleaner UX hook than expecting customers to find a per-brand checkbox.

A second decision: **ChatGPT Shopping returns answers from a different model (`gpt-5-3-mini`)** under `include.shopping: true` — its `visibility_score` is not comparable to a normal ChatGPT text response. Mixing those rows into the Insights aggregates would skew brand-level visibility numbers. The shopping data captured by Perplexity / AI Mode / Copilot is fine to keep in Insights because their shopping cards are a side-payload of the **same model's** answer; ChatGPT Shopping isn't.

### 1. Org-level "Shopping mode" toggle

- New field on `organizations` (e.g. `shopping_mode_enabled boolean not null default false`).
- New section in [`web/src/components/settings`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings) — Settings → Project (or its own card) — single checkbox "I'm in e-commerce / shopping". Default off. Admin-only via the existing `useUserRole().canAdmin` gate.
- New REST endpoint to read / update the flag (mirror the [`/api/settings/anthropic-key`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/api/settings/anthropic-key/route.ts) pattern).

### 2. Sidebar gating

- The Shopping entry in [`web/src/config/dashboard.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/config/dashboard.ts) is hidden unless `shopping_mode_enabled === true`. Combine the existing `requiredFeature: 'shopping_analytics'` plan gate with the new org-preference gate — both must pass.

### 3. Auto-bulk ChatGPT Shopping platform

When the org toggle flips from `false → true`:

[...]

PR #490, opened 2026-07-20
feat(insights): visibility rate as the headline metric across Insights

gkhngyk:
## Summary

The Insights headline was the average visibility score over ALL results. Every answer the brand is absent from contributes a 0, and tracking many platforms dilutes each appearance further — so the number reads near zero for most brands and buries the fact users act on: **how often the brand actually shows up**. This PR switches the whole visibility narrative to one prompt-level metric, end to end:

- **KPI row**: a new **Visibility Rate** card leads the row — distinct prompts the brand appeared in ÷ distinct prompts that produced results, under the same filters as the rest of the row (the Tracked Prompts KPI is the denominator on purpose, so the two cards always agree). Clicking it opens the existing per-model breakdown sheet, where the score average lives on. The old Visibility Score card is gone; Tracked Prompts / Mentions / Citations / Positive Sentiment are unchanged.
- **Leaderboard**: ranks and displays the same rate for the brand and every competitor, with an identical denominator for everyone (the #478 same-denominator rule kept — absence still costs; but a prompt counts once no matter how many platforms answered it). Each row shows `mentions · citations` plus the fraction behind its percentage (`appeared in X/Y prompts`), and the delta badge is the rate's point change vs the previous window.
- **Provider chart**: the Brand vs Competitors bars plot the same rate per provider instead of the score average, dividing everyone by the brand bucket's prompt count.

### Backend

Two additive RPC changes, both applied to the hosted DB:

- `visible_prompt_stats` (00027) — distinct prompts appeared / visible result rows / their score sum, mirroring `tracked_prompt_count`'s filters and the #155 shopping exclusion.
- `competitor_aggregates` (00028 + 00029) — distinct-prompt counts overall and per (model_used, platform) group, for the brand and each competitor. Payload shape is additive; existing readers are unaffected. Summing DISTINCT counts across two engines of one provider counts a shared prompt once per engine in numerator and denominator alike, so folded rates stay unbiased.

## Related issue

Follow-up to #457 (Tracked Prompts KPI) and #478 (same-denominator leaderboard rule); metric direction discussed with the team.

## Type of change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## How to test

1. Open `/dashboard/insights` — the first KPI card is **Visibility Rate** (a percentage), and the row has 5 cards; clicking the card opens the visibility breakdown sheet.
2. Change the date preset / model / region / topic filters — the rate, the leaderboard and the provider chart all move together and never disagree with the Tracked Prompts count.
3. **Leaderboard**: every row shows `N%`, a `mentions · citations` line and `appeared in X/Y prompts` where X/Y matches the displayed percentage; the brand's Y equals the Tracked Prompts KPI.
4. **Provider chart**: bars are percentages on the same scale as the leaderboard (e.g. a brand visible in 18 of 107 ChatGPT prompts shows ~17%).
5. Share of Voice keeps its own chart section below (no KPI card).
6. A brand with zero appearances in the window shows 0% without errors.

## Checklist

### DIFF
--- a/supabase/migrations/00027_visible_prompt_stats.sql
+++ b/supabase/migrations/00027_visible_prompt_stats.sql
@@ -1,0 +1,64 @@
+-- ── Visibility Rate KPI ───────────────────────────────────────────────────────
+--
+-- Headline metric change on Insights: the raw average visibility score over
+-- ALL results reads near zero for most brands (every answer the brand does
+-- not appear in contributes a 0), which buries the two numbers users act on:
+-- how OFTEN the brand shows up, and how GOOD it looks when it does. This RPC
+-- returns the "appeared" side of that split for the filtered window:
+--
+--   visible_prompts        distinct prompts with >= 1 answer mentioning or
+--                          citing the brand (numerator of Visibility Rate;
+--                          the denominator is tracked_prompt_count, 00024)
+--   visible_results        result rows where the brand appeared
+--   sum_visibility_visible sum of visibility_score over those rows, so the
+--                          caller derives "avg score when visible"
+--
+-- Filters mirror tracked_prompt_count / insights_aggregates exactly — same
+-- window, same #155 chatgpt-shopping exclusion — so the KPI row stays
+-- internally consistent.
+--
+-- SECURITY INVOKER (00014 convention): RLS on prompt_results/prompts scopes
+-- the caller to their own org's data.
+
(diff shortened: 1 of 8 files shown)


gkhngyk:
Added a refetch overlay on top of the rate work: on a date/filter change the content dims under a small floating spinner instead of silently showing the previous window's numbers until the new payload lands. First load keeps the skeleton; the filter bar stays interactive during the reload.

PR #512, opened 2026-07-23
fix(insights): divide Positive Sentiment by brand-mentioning results,…

### LINKED ISSUE #508: Insights: Positive Sentiment KPI is diluted by answers that never mention the brand
## Problem

The Insights **Positive Sentiment** KPI computes `positive_count ÷ ALL result rows` ([`web/src/lib/actions/tracking.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts), `positiveSentimentPct`). But the tracking pipeline short-circuits sentiment to `neutral` whenever the brand is absent from the answer ([`server/src/lib/response-parser.js`](https://github.com/ansvisor/ansvisor/blob/main/server/src/lib/response-parser.js)) — sentiment analysis never even runs on those rows.

So for a typical brand the denominator is dominated by answers where sentiment was never measured, and the card reads a permanent ~1–2%. Users read the remaining ~98% as "negative", when the actual negative count can be zero. Real production example over 30 days: 119 positive / 158 neutral-while-visible / **0 negative** among the 277 answers that actually mentioned the brand — i.e. **43% positive** — yet the card shows ~1% because 13,002 brand-absent rows sit in the denominator.

This is the same dilution family the headline Visibility Score had before #490 replaced it with the prompt-level rate.

## Fix

Make the KPI answer "when AI mentions the brand, how does it talk about it?":

- **Numerator**: `positive_count` — already returned by the `insights_aggregates` RPC. (Positive rows are by definition rows where the brand appeared, so the numerator needs no change.)
- **Denominator**: visible results — the count of rows with `mention_count > 0 OR citation_count > 0`. The Insights page already fetches this since #490: `visible_prompt_stats` returns `visible_results` (see `getVisibilityRateKpi` in [`web/src/lib/actions/tracking.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts)). No new RPC or migration needed.
- Guard the division (0 visible results → 0), same rounding style as the other KPI cards.
- Update the card's tooltip to say the percentage is **of answers that mention the brand**, e.g. "Percentage of AI answers mentioning your brand that described it positively."
- Check other consumers of `positiveSentimentPct` (`InsightsSummary` is reused by the topic detail page and reports) — they should either get the same corrected value automatically or be listed in the PR as intentionally unchanged.

## Acceptance criteria

- A brand with 119 positive answers out of 277 brand-mentioning answers shows ~43%, regardless of how many brand-absent answers exist in the window.
- Zero visible answers renders 0% without dividing by zero.
- The tooltip states the denominator (answers mentioning the brand).
- The KPI responds to the page's date/model/region/topic filters exactly like the neighboring cards.
- `yarn lint`, `yarn typecheck` and `yarn format:check` pass from `web/`.

## Getting started

See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. `npm run dev` with the seed data (`supabase/seed.sql`) is enough to verify. Comment here to claim the issue before opening a PR.

d180:
## Summary

The "Positive Sentiment" KPI on the Insights dashboard divided `positive_count`
by `total_results` — but sentiment analysis is skipped entirely for answers
that never mention the brand, so those rows sit in the denominator and can
never be positive. On brands with low mention volume this collapses the
score toward zero (e.g. a real 43% positive rate among brand-mentioning
answers displayed as ~1% once 13,002 brand-absent rows were mixed in).

This adds `mentioning_results` (`mention_count > 0 OR citation_count > 0`) to
the `insights_aggregates()` RPC and divides by that instead, with an explicit
zero-guard for brands that are tracked but never mentioned in the selected
window. Same fix covers the Insights dashboard, the Topics detail page, and
Reports (PDF + web), since they all read from `getInsightsSummary()`.

## Related issue

Closes #508

## Type of change

- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/supabase/migrations/00033_insights_sentiment_denominator.sql
+++ b/supabase/migrations/00033_insights_sentiment_denominator.sql
@@ -1,0 +1,79 @@
+-- #508 — Positive Sentiment KPI was dividing by total_results, which includes
+-- rows where the brand was never mentioned (sentiment analysis is skipped for
+-- those). Add mentioning_results so the app can divide by brand-mentioning
+-- answers only.
+CREATE OR REPLACE FUNCTION public.insights_aggregates(
+  p_brand_id   uuid,
+  p_platform   text         DEFAULT NULL,
+  p_models     text[]       DEFAULT NULL,
+  p_region     text         DEFAULT NULL,
+  p_date_from  timestamptz  DEFAULT NULL,
+  p_date_to    timestamptz  DEFAULT NULL,
+  p_prompt_id  uuid         DEFAULT NULL,
+  p_topic_id   uuid         DEFAULT NULL
+)
+RETURNS jsonb
+LANGUAGE sql
+STABLE
+SECURITY DEFINER
+SET search_path = public
+AS $$
+  WITH filtered AS (
+    SELECT pr.visibility_score, pr.mention_count, pr.citation_count,
+           pr.sentiment, pr.model_used, pr.created_at
+    FROM public.prompt_results pr
+    WHERE pr.brand_id = p_brand_id
+      AND pr.platform <> 'chatgpt-shopping'  -- #155 — isolate from Insights
+      AND (p_platform  IS NULL OR pr.platform    = p_platform)
+      AND (p_models    IS NULL OR pr.model_used  = ANY (p_models))
+      AND (p_region    IS NULL OR pr.region      = p_region)
+      AND (p_date_from IS NULL OR pr.created_at >= p_date_from)
+      AND (p_date_to   IS NULL OR pr.created_at <= p_date_to)
+      AND (p_prompt_id IS NULL OR pr.prompt_id   = p_prompt_id)
+      AND (p_topic_id  IS NULL OR EXISTS (
+             SELECT 1 FROM public.prompts p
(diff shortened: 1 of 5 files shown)


gkhngyk:
Solid fix — the RPC body is faithful to the current definition on main (only the new counter + JSON key added), the cur/prev sentiment windows use the same denominator, and going through `insights_aggregates` instead of `visible_prompt_stats` was the right call: it carries the `p_prompt_id` / `p_topic_id` filters the issue's suggested source lacks.

One mechanical blocker before this can land: #511 merged this morning and took migration numbers `00031`/`00032`, so this branch now conflicts on `supabase/schema.sql` and the migration number is taken. To resolve:

1. Rename the migration to `00033_insights_sentiment_denominator.sql`.
2. Rebase onto current `main` and regenerate the consolidated file with `bash supabase/build-schema.sh` — no need to hand-resolve the `schema.sql` conflict, the script rebuilds it from the migration files.
3. Push; the schema CI check will confirm.

Will approve as soon as the rebase is up. (Note for after merge: the migration gets applied to the hosted project by a maintainer — nothing you need to do there.)

d180:
@gkhngyk done the rebase and the needed changes.

PR #513, opened 2026-07-23
feat(mcp): expose topic suggestions via list/accept/dismiss tools

### LINKED ISSUE #498: MCP: expose topic suggestions (list / accept / dismiss)
## Problem

#463 (PR #497) adds persisted AI topic suggestions to the Topics page, but the MCP surface has no visibility into them. An agent can call `list_topics` and `list_prompts`, yet cannot answer "what topics is the platform suggesting I add?" or act on a suggestion — even though the rows sit in `topic_suggestions`.

**Depends on #497** — the `topic_suggestions` table and the accept/dismiss flows land there. Start after it merges.

## Design

Add three tools to the MCP server, following the existing patterns in [`web/src/lib/mcp/server.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/mcp/server.ts) (tool registration + Zod input schema) and [`web/src/lib/mcp/data.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/mcp/data.ts) (auth-scoped data helpers):

1. **`list_topic_suggestions`** `(brand_id)` — pending (`status = 'new'`) suggestions: `id`, `name`, `reason`, `generated_at`. Read-only; mirrors `list_topics`.
2. **`dismiss_topic_suggestion`** `(suggestion_id)` — sets `status = 'dismissed'`. Dismissed names never reappear in future generations (handled server-side already).
3. **`accept_topic_suggestion`** `(suggestion_id)` — creates **exactly one** topic from the suggestion's name and marks the suggestion `status = 'added'` with `added_topic_id` set. Reference flow: `acceptTopicSuggestion` in [`web/src/lib/actions/topic-suggestions.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic-suggestions.ts). Use the **singular** `createTopic` action semantics — a single insert into `topics`; the plural `createTopics` replaces the brand's whole topic list and must not be used here. `update_opportunity_status` is the pattern reference for a write tool.

Brand/suggestion access must be scoped to the caller like every other tool in `data.ts` (a suggestion belonging to another org returns not-found, not data).

**Out of scope**: triggering generation (refresh) from MCP. It runs a search-grounded LLM call and is plan-gated; the Topics page button covers it. A read/act-only MCP surface is the right first step.

## Acceptance criteria

- `list_topic_suggestions` returns only `status='new'` rows for brands the caller can access.
- `accept_topic_suggestion` creates one topic, leaves existing topics untouched, and moves the suggestion to `added` with `added_topic_id` set; accepting the same suggestion twice returns an error, not a duplicate topic.
- `dismiss_topic_suggestion` moves the suggestion to `dismissed`; it no longer appears in `list_topic_suggestions`.
- Tool descriptions explain when an agent should reach for each tool (see existing tools for tone).

## Getting started

See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. From `web/`, `yarn lint`, `yarn typecheck` and `yarn format:check` must pass. Comment here to claim the issue before opening a PR.

d180:
## Summary

Agents connected via MCP could list topics and prompts but had no way to see
or act on the AI-generated topic suggestions added in #497 — accepting or
dismissing one required switching to the web dashboard.

This adds three MCP tools: `list_topic_suggestions`, `accept_topic_suggestion`,
and `dismiss_topic_suggestion`, plus matching REST twins under
`/api/mcp/topic-suggestions/*` (mirroring the pattern every other MCP
capability in this codebase already follows). Accept creates exactly one
topic via a singular insert (never a bulk replace) and both accept/dismiss
are scoped to `status = 'new'`, so acting on an already-decided suggestion
returns not-found instead of silently double-processing or creating a
duplicate topic. Generating new suggestions stays out of scope — that
remains plan-gated behind the UI's refresh button.

Verified end-to-end against a local Supabase instance: correct org isolation
(wrong-org calls return not-found), and a real MCP client (Claude Code)
connected over the Streamable HTTP endpoint successfully listed and accepted
a live suggestion.

## Related issue

Closes #498

## Type of change

- [x] `feat` — New feature

## Checklist

### DIFF
--- a/web/src/app/api/mcp/topic-suggestions/[id]/accept/route.ts
+++ b/web/src/app/api/mcp/topic-suggestions/[id]/accept/route.ts
@@ -1,0 +1,35 @@
+import { NextResponse } from 'next/server';
+import { authenticateMcpRequest } from '@/lib/mcp-auth';
+import { acceptTopicSuggestionFor } from '@/lib/mcp/data';
+
+/**
+ * POST /api/mcp/topic-suggestions/[id]/accept
+ *
+ * Parallel REST surface for the `accept_topic_suggestion` MCP tool — same
+ * data-layer function, same ownership guarantee, same status = 'new' guard
+ * that keeps a repeated accept from creating a second topic.
+ */
+export async function POST(req: Request, { params }: { params: Promise<{ id: string }> }) {
+  const auth = await authenticateMcpRequest(req);
+  if (auth instanceof NextResponse) return auth;
+
+  const resolvedParams = await params;
+  const suggestionId = resolvedParams.id;
+  if (!suggestionId) {
+    return NextResponse.json({ error: 'id is required' }, { status: 400 });
+  }
+
+  try {
+    const result = await acceptTopicSuggestionFor(auth, suggestionId);
+    if (!result) {
+      return NextResponse.json(
+        { error: 'Suggestion not found, not pending, or already processed' },
+        { status: 404 },
+      );
+    }
+    return NextResponse.json(result);
+  } catch (err) {
+    const message = err instanceof Error ? err.message : 'Failed';
+    return NextResponse.json({ error: message }, { status: 500 });
+  }
+}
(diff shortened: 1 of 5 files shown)


gkhngyk (approved):
Clean work — all four acceptance criteria met, org isolation done right. Thanks!

PR #515, opened 2026-07-23
fix(reports): lead the KPI section with the visibility rate, with a fallback for old snapshots

### LINKED ISSUE #492: Reports: lead the KPI section with the visibility rate, with a fallback for old snapshots
## Problem

Reports snapshot their data as a frozen JSONB payload ([`web/src/lib/actions/reports.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/reports.ts)) and the viewer ([`web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/reports/%5Bid%5D/page.tsx)) still leads its KPI section with `payload.insights.avgVisibilityScore` — the all-rows score average the dashboard just moved away from (#490). A customer reads "Visibility ~1%" in a report and a meaningful rate on the Insights page for the same period.

## Fix

**Snapshot (report generation):**

- When the `kpis` section is enabled, also fetch the Visibility Rate the dashboard shows — `getVisibilityRateKpi` + `getTrackedPromptsKpi` from [`web/src/lib/actions/tracking.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts) (same date range) — and store `visibilityRate: { visiblePrompts, promptCount, ratePct }` in the payload.
- The competitors section needs no generation change: it already stores `CompetitorComparisonEntry[]`, which carries `visibilityRate` / `visiblePrompts` / `promptCount` since #490.

**Viewer:**

- KPI row: lead with the rate (`X%`, with the `visiblePrompts/promptCount` fraction as its sub-line); keep the score average as a secondary stat.
- Competitors table: show each entry's `visibilityRate` with `appeared in X/Y prompts`, mirroring the dashboard leaderboard.
- The PDF export ([`web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/_report-pdf.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/reports/%5Bid%5D/_report-pdf.tsx)) renders the same KPI and competitor numbers — apply the identical rate-first display and fallback there.
- **Old snapshots must keep rendering**: payloads generated before this change have neither `visibilityRate` nor the new entry fields. Fall back to the current display (`avgVisibilityScore`) whenever the new fields are absent — no migration or backfill of stored reports.

**Out of scope:** the trend chart and the per-prompt / per-topic performance tables stay score-based — per-prompt and per-topic averages are meaningful at that granularity. The server-side executive-summary prompt may mention "visibility score"; adjusting its wording can ride along if trivial, otherwise skip.

## Acceptance criteria

- A newly generated report's KPI section leads with the same visibility rate the Insights page shows for the identical date range, and its competitors table matches the dashboard leaderboard's numbers.
- A report generated before this change still renders exactly as today (no crashes, no empty KPI).

## Getting started

Depends on #490 (the entry fields and `getVisibilityRateKpi`) — start after it merges. See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. Comment here to claim the issue before opening a PR.

d180:
## Summary

Reports still snapshotted and rendered the old all-rows `avgVisibilityScore` after the Insights dashboard moved to the prompt-level Visibility Rate in #490, so a report's KPI section could show a wildly different number than the Insights page for the identical period (e.g. 59.7% vs. 100%). This snapshots the Visibility Rate at report-generation time and leads the KPI row and competitors table with it in both the on-screen viewer and the PDF export — the score average stays as a secondary stat. Reports generated before this change keep rendering exactly as before (no `visibilityRate` in their stored payload → automatic fallback to the old display). Also tightens the AI executive-summary prompt so its prose narrates the rate, not the score, as the headline number, and doesn't fabricate a delta for a metric that has no previous-period value in the snapshot.

## Related issue

Closes #492

## Type of change

- [x] `fix` — Bug fix

## Checklist


gkhngyk (approved):
Excellent work — snapshot sourced from the same RPC pair as the Insights KPI so the numbers match by construction, the runtime field-check for pre-#490 competitor snapshots is exactly the right call over trusting the type, and the no-fabricated-delta instruction in the summary prompt shows real care. Thanks!

Issue #514, opened 2026-07-23
Reports: AI executive summary misreads mentions/citations % change as absolute counts


d180:
## Problem

`InsightsSummary.mentionsChange` and `InsightsSummary.citationsChange` (`web/src/lib/actions/tracking.ts`) are **percentage** deltas vs. the previous period of equal length — e.g. `1262` means "+1262%", matching how the KPI card renders it (`+1262%`, see the `Delta` component's `%` suffix in both the report viewer and PDF).

The report's AI executive summary (`server/src/routes/reports.js`, `POST /api/reports/summary`) receives the raw JSON snapshot and narrates these fields as if they were absolute counts. Example output for a report where `totalMentions: 354` and `mentionsChange: 1262`:

> "total mentions increasing by 1,262 to reach 354"

That's nonsensical — a literal reading implies a previous value of `354 - 1262 = -908`, which is impossible. The same happened with citations (`totalCitations: 177`, described as "citations climbing by 1,080, totaling 177").

Found while verifying #492 — unrelated to that fix (which only concerns the visibility-rate headline metric), but real and reproducible with any report whose mentions/citations grew a lot from a small prior base.

## Fix

In `SUMMARY_SYSTEM_PROMPT` (`server/src/routes/reports.js`), clarify that `mentionsChange`/`citationsChange` (and any other `*Change` field sourced from `getInsightsSummary`) are **percentage** deltas, not absolute counts, and instruct the model to phrase them accordingly (e.g. "mentions rose 1,262%, to 354" rather than "increased by 1,262 to reach 354").

## Acceptance criteria

- Regenerating a report where mentions/citations grew sharply from a small base produces a summary that states the change as a percentage, consistent with the KPI card's `+X%` display, and never implies an impossible previous value.

gkhngyk:
Confirmed — the math in the example checks out (prev ≈ 26 → 354 is +1,262%), and the system prompt indeed never states these fields are percentages. Good catch and a well-scoped fix.

[...]

PR #521, opened 2026-07-23
fix: handle zero-base insight deltas

### LINKED ISSUE #520: Insights & Reports: *Change deltas silently switch to absolute counts on a zero base but always render with a % suffix
## Problem

`InsightsSummary.mentionsChange` / `citationsChange` (`web/src/lib/actions/tracking.ts`) are percentage deltas **except** when either window is zero, where the formula falls back to an absolute difference:

```ts
mentionsChange =
  curMentions > 0 && prevMentions > 0
    ? Math.round(((curMentions - prevMentions) / prevMentions) * 100)
    : curMentions - prevMentions;
```

Every consumer renders the value with an unconditional `%` suffix — the Insights KPI cards (`DeltaBadge`), the report viewer's `Delta`, and the PDF's `DeltaText`. So a brand going 0 → 354 mentions shows **"+354%"** in three surfaces, when the truth is "+354 mentions (new)". The mixed semantics also can't be narrated correctly by the report's AI summary (#514 fixes the prompt for the common percentage case, but the model cannot know which meaning applies to a given value).

## Fix direction

Make the field's meaning unambiguous instead of guessing at render time. The least invasive shape:

- Keep `*Change` as the percentage delta and set it to `null` when there is no meaningful base (prev = 0), **and**
- Expose the raw previous-period counts (`prevMentions`, `prevCitations`) on `InsightsSummary` so consumers can render the zero-base case honestly (e.g. a "new" badge or "+354" without the % suffix) and the report snapshot carries values the AI summary can narrate safely.

Alternative shapes are fine if they keep one meaning per field — the requirement is that no consumer can render an absolute count with a % suffix.

Consumers to update together (all read the same struct, so they must land in one PR): Insights KPI cards, report viewer `Delta`, report PDF `DeltaText`, and the report summary prompt's description of the fields (coordinate with #514). Old report snapshots store the current mixed-meaning values — the viewer/PDF fallback for them should keep today's rendering rather than reinterpreting immutable data.

## Acceptance criteria

- A brand with prev = 0 and cur > 0 never shows a `%` suffix for that delta on Insights, the report viewer, or the PDF.
- Percentage deltas keep rendering exactly as today when both windows are non-zero.
- Reports generated before the change render unchanged.
- `yarn lint`, `yarn typecheck` and `yarn format:check` pass from `web/`.

## Related

- #514 (report summary prompt misreads the percentage case — stays scoped to the prompt)

## Getting started

See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md) for setup. Comment here to claim the issue before opening a PR.

abdullah91111:
## Summary

Fixes ambiguous mentions and citations change deltas when the previous-period count is zero.

- Keeps `mentionsChange` and `citationsChange` strictly percentage-based
- Returns `null` when no meaningful percentage can be calculated from a zero base
- Adds `prevMentions` and `prevCitations` to distinguish zero-base growth from unavailable comparison data
- Displays zero-base growth as `+N new` instead of an incorrect `+N%`
- Updates the Insights dashboard, report viewer, and report PDF together
- Preserves the existing rendering of reports generated before this change
- Clarifies delta semantics for the AI report-summary prompt
- Adds regression tests for zero-base growth and drops to zero

## Related issue

Closes #520

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

## Testing

- Web: 45 tests passed across 10 test files
- Server: 123 tests passed across 10 test files
- All changed files pass Prettier
- `git diff --check` passes

The repository-wide format check reports existing CRLF formatting differences in untouched files in the Windows checkout. All files modified by this PR pass Prettier individually.


gkhngyk (changes requested):
This matches #520's fix direction almost exactly — the single-meaning `percentageChange` helper, the raw prev counts, all four consumers updated together, and the `Object.hasOwn` fallback that keeps pre-existing snapshots rendering unchanged is exactly right. The prompt paragraph is accurate too (and effectively covers #514's ask — worth cross-linking there so effort isn't duplicated).

One real blocker, invisible to CI:

**`export function percentageChange` is a sync export from a `'use server'` module.** `tracking.ts` opens with `'use server'`, and Next.js only allows **async function** exports from such modules — `next build` fails with *"Only async functions are allowed to be exported in a 'use server' file."* Nothing catches it today: CI runs prettier/eslint/tsc/vitest but no `next build`, the Vercel build is canceled on PR branches by the ignored-build-step, and vitest doesn't enforce the directive. Merged as-is, this breaks the production build on main.

Fix: move `percentageChange` into a plain module (e.g. `web/src/lib/metrics.ts`), import it from `tracking.ts`, and point the test at the new module.

Tiny nit while you're in there: `CountDeltaBadge` uses `text-green-600` where the report viewer uses `emerald-600` — matching them keeps the delta color consistent across surfaces.

Everything else is good to go — happy to approve as soon as the helper moves.

abdullah91111:
alright will fix it

gkhngyk (approved):
Verified the fix end-to-end: `percentageChange` now lives in `web/src/lib/metrics.ts` (plain module), `tracking.ts` is back to async-only exports, the test moved with it, and the emerald nit is fixed too. I also ran a full `next build` against this branch locally — it compiles and completes cleanly, so the production-build blocker is definitively gone. Thanks for the quick, thorough turnaround!

PR #558, opened 2026-07-29
fix: preserve YouTube video IDs in citation URLs

### LINKED ISSUE #557: Citations: query-keyed URLs (YouTube watch) lose their identity — broken links and merged counts
## Problem

Citation URL aggregation normalizes URLs by stripping the query string and fragment (to dedupe `?utm_…` noise) before bucketing:

- [web/src/lib/actions/citations.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts) — the overview's URL aggregation (`parsed.search = ''`)
- [web/src/lib/actions/tracking.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts) — `getPromptDetail`'s Top Sources URL aggregation keeps its own copy of the same rule

For sites whose page identity lives in the query string this is wrong. YouTube is the visible case: `https://www.youtube.com/watch?v=<id>` normalizes to `https://www.youtube.com/watch`, so:

1. **Broken links** — clicking a YouTube row on the Citations page (or the title link on the per-URL detail page from #535) opens the bare `/watch` URL instead of the video.
2. **Merged counts** — every distinct YouTube video the brand is cited in collapses into one `youtube.com/watch` row; its citation totals are the sum across different videos, and the per-URL detail page aggregates them as if they were one page.

(`youtu.be/<id>` and `/shorts/<id>` are path-keyed and unaffected.)

## Fix

Make the normalization host-aware instead of blanket-stripping:

1. Move the URL normalization helper (and the loose host+path match key) into a plain lib module, e.g. `web/src/lib/citations/normalize.ts`, and import it from both `citations.ts` and `tracking.ts` — the two copies exist today and have already drifted once. (Note: it can't be exported from `citations.ts` directly — that file is `'use server'`, which only allows async action exports.)
2. Keep the general rule (drop query + fragment, strip one trailing slash), but preserve identity parameters for known query-keyed hosts. v1 list: `youtube.com` / `m.youtube.com` path `/watch` → keep `v`. Structure it as a small host→params table so future cases are one-line additions.
3. No migration needed: all aggregations are computed at read time, so historical results re-bucket correctly as soon as the rule changes.

## Acceptance

- A citation of `youtube.com/watch?v=abc` links to the actual video from the Citations page, the prompt detail Top Sources tab and the per-URL detail page.
- Two different videos produce two distinct URL rows with separate counts; `?v=abc` with extra `utm_…` params still buckets with plain `?v=abc`.
- Non-query-keyed URLs bucket exactly as before.

abdullah91111:
## Summary

Fixes citation URL aggregation for query-keyed YouTube watch URLs.

- Preserves the `v` parameter for `youtube.com/watch` and `m.youtube.com/watch`.
- Removes unrelated query parameters and fragments, including UTM parameters.
- Prevents different YouTube videos from being merged into one citation row.
- Uses one shared normalization helper across Citations, per-URL details, and prompt Top Sources.
- Adds regression tests for YouTube and existing non-query-keyed behavior.

## Related issue

Closes #557

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/lib/actions/citations.ts
+++ b/web/src/lib/actions/citations.ts
@@ -11,6 +11,7 @@ import {
   SOURCE_CATEGORIES,
 } from '@/lib/citations/classify';
 import { classifyArticleType, type ArticleType } from '@/lib/citations/article-type';
+import { citationUrlMatchKey, normalizeCitationUrl } from '@/lib/citations/normalize';
 
 // ─── Types ────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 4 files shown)


gkhngyk (approved):
Thanks for the clean fix — tests included and all green locally as well. 🙌

PR #559, opened 2026-07-29
fix(web): show error state instead of empty state on invoice fetch failure

### LINKED ISSUE #550: Billing invoice list silently swallows API errors — customers with invoices can be shown "No invoices"
## Problem

[web/src/components/settings/invoice-list.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/invoice-list.tsx) fetches `/api/stripe/invoices` with no error handling:

```ts
fetch('/api/stripe/invoices')
  .then((r) => r.json())
  .then((data) => {
    if (Array.isArray(data)) setInvoices(data);
  })
  .finally(() => setLoading(false));
```

If the route returns a 500 or an `{ error }` body, the `Array.isArray` guard silently swallows it and `.finally` clears `loading` — so a customer with real invoices is shown the **"No invoices"** empty state. That's the worst kind of silent failure: indistinguishable from a legitimately empty list. There's also no `.catch()`, so a network failure becomes an unhandled promise rejection.

## Fix

Every sibling settings component ([plan-switcher.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/plan-switcher.tsx), [api-keys-section.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/api-keys-section.tsx), [agent-section.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/agent-section.tsx)) already uses the `if (!res.ok) throw new Error(body.error || …)` pattern — this file is the one straggler.

1. Check `res.ok` and throw/parse the error body.
2. Add a `.catch()` that sets an `error` state.
3. Render a distinct error state (with a retry button) separate from the "No invoices" empty state.

## Acceptance

- When the invoices API fails, the user sees an error state with a retry option — never the "No invoices" empty state.

arambu1a:
Closes #550

## Summary
- `invoice-list.tsx` had no error handling on its fetch to `/api/stripe/invoices` — a failed request silently rendered "No invoices yet," indistinguishable from a real empty list
- Now checks `res.ok`, parses the error body, and renders a distinct error state with a retry button, matching the pattern in `agent-section.tsx`

## Test plan
- [x] Verified locally: forced a module-load-time Stripe crash (HTML 500) — error state renders correctly, not the empty state
- [x] Verified locally: forced a route-level JSON error (`{error: "..."}`, 500) — error message renders correctly, retry button works

### DIFF
--- a/web/src/components/settings/invoice-list.tsx
+++ b/web/src/components/settings/invoice-list.tsx
@@ -1,8 +1,8 @@
 'use client';
 
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useTranslations } from 'next-intl';
-import { Download } from 'lucide-react';
+import { Download, AlertCircle, RotateCw } from 'lucide-react';
 import {
   Table,
   TableBody,
@@ -12,6 +12,7 @@ import {
   TableRow,
 } from '@/components/ui/table';
 import { Badge } from '@/components/ui/badge';
+import { Button } from '@/components/ui/button';
 import { Skeleton } from '@/components/ui/skeleton';
 
 interface Invoice {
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Thanks — clean fix, matches the sibling settings components and the non-JSON 500 case is handled nicely. One small ask: could you add `Closes #550` to the PR description so the issue auto-closes on merge?

PR #566, opened 2026-07-30
fix(web): return invite/accept errors as values instead of throwing

### LINKED ISSUE #548: Team invitations: inviteMember / acceptInvitation throw — users see the masked digest error instead of actionable messages
## Problem

Two server actions in [web/src/lib/actions/team.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/team.ts) throw on common, user-reachable failures. In production a thrown server action surfaces as the masked `An error occurred in the Server Components render… digest: …` string, so the actionable message is lost.

**`inviteMember`** throws:
- `Invalid email address`
- `This user is already a member of your team`
- `A pending invitation already exists for this email`
- Supabase insert errors

Repro: invite the same email twice from Settings → Team. The second attempt shows the digest string instead of "A pending invitation already exists for this email".

**`acceptInvitation`** throws:
- `Invitation not found` / `This invitation is no longer valid` / `This invitation has expired`
- ``This invitation was sent to ${inviteEmail}. Please sign in with that email.``

Repro: open an invitation link while signed in with a different account. The user hits a dead end with a digest error instead of being told to switch accounts — the single most actionable message in the flow.

## Fix

Follow the pattern from PR #536 (`addPromptToBrand` in [web/src/lib/actions/prompt.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/prompt.ts) is the in-repo reference implementation): return errors as values instead of throwing.

1. `inviteMember` → return `{ inviteLink, emailSent } | { error: string }`; branch on `'error' in result` in `handleSubmit` of [web/src/components/settings/team-section.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/team-section.tsx).
2. `acceptInvitation` → extend its existing result type with `{ error: string }` and render the error **inline** in [web/src/components/team/accept-invitation-card.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/team/accept-invitation-card.tsx) (inline beats a toast here — the message tells the user what to do next).

## Acceptance

- Duplicate invite shows the pending-invitation message.
- Accepting with the wrong account shows the "sign in with that email" message inline on the card.

arambu1a:
## Summary
- `inviteMember` and `acceptInvitation` threw on common, user-reachable failures (duplicate invite, wrong account, expired token, invalid email, insert errors), which production masks as a digest string instead of showing the actionable message.
- Both now return `{ error: string }` instead of throwing, following the pattern from #536 (`addPromptToBrand`).
- `team-section.tsx` branches on the invite result and toasts the error.
- `accept-invitation-card.tsx` branches on the accept result and renders the error inline (inline beats a toast here since it tells the user what to do next, e.g. which account to sign in with).

Closes #548 

## Test plan
- [x] `tsc --noEmit` passes
- [x] `npm run lint` passes
- [x] Manually verify: inviting the same email twice shows "A pending invitation already exists for this email"
- [x] Manually verify: accepting an invite link while signed in with a different account shows the "sign in with that email" message inline on the card

### DIFF
--- a/web/src/components/settings/team-section.tsx
+++ b/web/src/components/settings/team-section.tsx
@@ -433,10 +433,14 @@ function InviteDialog({ canInvite, onInvited }: { canInvite: boolean; onInvited:
     }
     setSubmitting(true);
     try {
-      const { inviteLink: link, emailSent: sent } = await inviteMember(email.trim(), role);
-      setInviteLink(link);
-      setEmailSent(sent);
-      toast.success(sent ? 'Invitation sent' : 'Invite created — share the link');
+      const result = await inviteMember(email.trim(), role);
+      if ('error' in result) {
+        toast.error(result.error);
+        return;
+      }
+      setInviteLink(result.inviteLink);
+      setEmailSent(result.emailSent);
+      toast.success(result.emailSent ? 'Invitation sent' : 'Invite created — share the link');
       onInvited();
     } catch (error) {
       toast.error(error instanceof Error ? error.message : 'Failed to send invitation');
--- a/web/src/components/team/accept-invitation-card.tsx
+++ b/web/src/components/team/accept-invitation-card.tsx
@@ -47,9 +47,11 @@ export function AcceptInvitationCard({
   const [confirmPassword, setConfirmPassword] = useState('');
   const [isAccepting, setIsAccepting] = useState(false);
   const [isSwitching, setIsSwitching] = useState(false);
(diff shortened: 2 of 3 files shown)


gkhngyk (approved):
Thanks — solid work again. The inline error on the accept card (instead of a toast) is exactly the right call. 🙌

Issue #464, opened 2026-07-17
Topics: leaderboard numbers disagree with Insights — unpaginated scan, no shopping exclusion, fixed window


gkhngyk:
## Problem

The Topics page's mentions/citations/visibility numbers don't match the Insights page for the same brand. Verified on a production brand — three stacked causes, largest first:

1. **Silent 1000-row cap.** `getTopicsOverview` ([`web/src/lib/actions/topic.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic.ts)) fetches the 30-day `prompt_results` window in a single un-paginated select. The verified brand has **13,352 rows** in that window; PostgREST silently returns the first 1000 — and the query has **no `ORDER BY`**, so every topic aggregate is computed from an arbitrary ~7.5% sample. Same defect family as #427/#430/#450. Insights uses `.range()` + exact counts and is correct.
2. **No `chatgpt-shopping` exclusion.** Insights excludes shopping from every aggregate (#155); Topics includes it. The verified brand has **1,384 shopping rows** in the window → +133 mentions / +35 citations of divergence on its own. (On this brand they're pre-#432 residue that ages out of the window by mid-August — the leak itself is plugged, last shopping row 2026-07-15, prompts clean — but for shopping-enabled brands the divergence is permanent.)
3. **Window mismatch.** Topics hardcodes 30 days; Insights follows the user's date preset. Even after fixing 1–2 the numbers only line up when Insights is on 30d, and nothing in the Topics UI says which window it shows.

Bonus: the same function's `prompts` select fetches **all** org prompts un-paginated and filters to the brand client-side — fine at current scale, same 1000-row trap for large orgs.

## Fix

In `getTopicsOverview`:

- Add `.neq('platform', 'chatgpt-shopping')` to the results query, matching Insights (#155).
- Page the results scan with deterministic ordering (`created_at, id`) and a hard row ceiling, aggregating per batch — the `scanFilteredResults` pattern in [`web/src/lib/actions/citations.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts).
- Scope the prompts query to the brand at the DB level (join/filter on the brand's prompt sets) instead of fetching the whole org and filtering in JS; page it for safety.
- UI: label the window ("Last 30 days") on the Topics page header/leaderboard so the fixed window is explicit.

## Acceptance criteria

[...]

PR #568, opened 2026-07-30
fix(reports): rank prompt/topic performance by visibility rate, not diluted average

### LINKED ISSUE #564: Reports: prompt/topic performance still rank by the zero-diluted average score (shown with a % sign)
## Problem

#562 switched the All Prompts **Visibility** column to the prompt-level **Visibility Rate** because averaging `visibility_score` across **all** runs lets zero-visibility runs dilute heavily-mentioned prompts into misleading single digits. The Reports feature still uses the old formula in two sections — and renders the 0–100 **score** with a `%` sign, as if it were a percentage:

- [`getPromptPerformance`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/reports.ts) (`web/src/lib/actions/reports.ts`, ~line 208) — ranks best/worst prompts by `avg(visibility_score)` over all runs
- [`getTopicPerformance`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/reports.ts) (~line 334) — same all-runs average per topic, with a points change vs the previous window
- Rendering: [report detail page](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/reports/%5Bid%5D/page.tsx) (`{tp.avgVisibility}%`, `{p.avgVisibility}%`) and [`_report-pdf.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/reports/%5Bid%5D/_report-pdf.tsx) (lines ~215, ~463)

There is also an adjacent correctness bug in the same two functions: their `prompt_results` selects are **un-paginated**, so PostgREST silently caps them at 1000 rows (the #430/#464 defect family — the comment above `getPromptVisibilitySummaries` in `tracking.ts` documents this exact hazard). Busy brands get best/worst lists computed from a truncated window. `getTopicPerformance` spans two windows in one select, so it hits the cap even sooner.

## Fix

1. **`getPromptPerformance`**: also select `citation_count`; page the scan at 1000 rows/page (the paged-loop pattern already exists later in the same file for the citation-evidence scan); accumulate `visibleRuns` (rows with `mention_count > 0 || citation_count > 0`); add `visibilityRate` to `ReportPromptPerf` — `Math.round((visibleRuns / runs) * 1000) / 10`, the exact rounding used in `getPromptVisibilitySummaries` (#562); rank best/worst by the rate (tie-break by mentions). Keep `avgVisibility` in the payload.
2. **`getTopicPerformance`**: same — per-window `visibleRuns`, topic `visibilityRate`, and `change` becomes rate points vs the previous window. Page the scan here too. Keep `avgVisibility`.
3. **Rendering** (detail page + PDF): show the rate with the `%` sign (it genuinely is a percentage now). **Old stored reports must keep rendering**: reports are immutable snapshots, so payloads generated before this change won't have `visibilityRate` — fall back to the existing `avgVisibility` cell (without adding a misleading `%`, e.g. render `NN` or `NN/100` for legacy payloads). The competitor table on the detail page already does exactly this dance for its own rate switch (`hasCompetitorVisibilityRate ? c.visibilityRate : c.avgVisibilityScore`) — mirror it.
4. No migration needed — reports compute at creation time and store the payload.

## Acceptance

- A newly created report ranks best/worst prompts and topics by visibility rate and labels the values as percentages.
- Reports created before the change still render without errors, showing their stored score values.
- A brand with more than 1000 results in the report window gets numbers computed from the full window, not a truncated page.

arambu1a:
## Summary
- `getPromptPerformance` and `getTopicPerformance` now rank/report by prompt-level **Visibility Rate** (share of runs with a mention or citation) instead of an all-runs average `visibility_score`, which let zero-visibility runs dilute heavily-mentioned prompts/topics into misleading single digits — the same class of bug `#562` fixed for the All Prompts tab. Rate uses the same rounding as `getPromptVisibilitySummaries`.
- Both scans were un-paginated, so PostgREST silently capped them at 1000 rows on busy brands (the `#430`/`#464` defect family). Both now page in 1000-row batches, same pattern as the existing citation-evidence scan.
- Report detail page and PDF render the rate with a `%` sign, but fall back to the legacy `avgVisibility/100` cell for reports generated before this shipped (reports are immutable snapshots) — mirroring the competitor table's existing `visibilityRate` fallback dance.
- Updated the AI executive-summary system prompt so generated prose also treats `avgVisibility` as a raw score, not a percentage, and prefers `visibilityRate` when present.

Closes #564

## Test plan
- [x] `yarn lint` passes
- [x] `yarn typecheck` passes
- [x] `yarn format:check` passes
- [x] Manually generate a new report and confirm best/worst prompts and topics are ranked by visibility rate and labeled with `%`
- [x] Open an older stored report and confirm it still renders without errors, showing `NN/100` (no `%`) for legacy payloads

### DIFF
--- a/server/src/routes/reports.js
+++ b/server/src/routes/reports.js
@@ -24,6 +24,8 @@ const SUMMARY_SYSTEM_PROMPT = `You are an AI search visibility analyst writing t
 
 You will receive a JSON snapshot of the brand's metrics for the report period. Depending on the report's sections it may include: overall visibility KPIs (with deltas vs the previous period), a daily visibility trend, share of voice, a competitor comparison, per-topic performance, best/weakest prompts, observed query fan-outs, AI-referred traffic, shopping visibility, a site audit score, and a citations overview. Every delta in the snapshot compares against the previous period of equal length.
 
+In promptPerformance and topicPerformance, visibilityRate is the percentage of runs in which the brand appeared at all — that is the number to quote for a prompt or topic, and topicPerformance.change is a percentage-point change in that rate. Their avgVisibility is a 0-100 intensity score, not a percentage: never write it with a % sign, and prefer visibilityRate whenever it is present.
+
 For insights KPIs, mentionsChange and citationsChange are percentage changes, never raw counts. A null change with prevMentions or prevCitations equal to 0 means there is no meaningful percentage base: describe the current total as new growth from zero (for example, "354 new mentions"), not as a percentage. A null previous count means comparison data is unavailable. visibilityChange and sentimentChange are percentage-point changes.
 
 Write a 1-2 paragraph executive summary in English for a marketing executive. It must tell the CHANGE STORY of the period, not describe a static snapshot:
--- a/web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/_report-pdf.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/reports/[id]/_report-pdf.tsx
@@ -201,18 +201,28 @@ function HBar({ label, pct, value }: { label: string; pct: number; value: string
 }
 
 function PromptTable({ title, prompts }: { title: string; prompts: ReportPromptPerf[] }) {
+  // Reports generated before the rate switch stored an all-runs average score
+  // (0-100 intensity, not a percentage) — render those unchanged, without a
+  // misleading % sign.
+  const hasVisibilityRate = prompts.some((p) => typeof p.visibilityRate === 'number');
   return (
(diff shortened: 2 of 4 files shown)


gkhngyk (approved):
Thanks — recipe followed to the letter, and updating the LLM summary prompt so it quotes the rate correctly was a genuinely sharp catch beyond the issue's scope. 🙌

PR #582, opened 2026-08-02
feat: trigger prompt volume analysis after checkout

### LINKED ISSUE #541: Auto-run prompt volume analysis after checkout — drop the manual Analyze step for new users
## Problem

After onboarding, a new user's prompts start tracking automatically once payment completes — but **volume & competition data** still requires them to find the Prompts page and click *Analyze prompts* manually. Most new users don't discover this, so the Volume/Competition columns sit empty.

## Proposal

Kick off the volume analysis automatically right after checkout, reusing the exact pattern the post-payment flow already uses for tracking.

Current post-payment flow ([web/src/app/api/stripe/success/route.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/api/stripe/success/route.ts)): validate session → update org subscription → `alignPromptsToPlanForOrg` → call Express `POST /api/internal/trigger-tracking` with `CRON_SECRET` auth.

**Changes:**

1. **Extract the analysis core into a shared lib function.** Pull the per-prompt loop out of `POST /api/volumes/analyze-batch` ([server/src/routes/volumes.js](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/volumes.js)) — intent/keyword extraction (`extractIntentKeywords`) + DataForSEO fetch (`fetchAndSaveVolumes`), reusing saved keywords when present — so the manual button and the auto-run share one implementation.
2. **New internal endpoint** `POST /api/internal/analyze-volumes` (`CRON_SECRET` bearer auth, alongside the existing `/api/internal/*` endpoints). Body: `{ brandId }`. It must **respond immediately (202)** and process in the background — analyzing ~50 prompts takes minutes and must never block the caller. Results land in `prompt_volumes` as they complete, so the Prompts page fills in on its own.
3. **Fire-and-forget call from the Stripe success route**, right after the `trigger-tracking` call. Any failure is logged and never affects the checkout redirect. (Optionally mirror in the `checkout.session.completed` webhook branch for checkouts that never hit the redirect — must be idempotent if both fire.)

**Free of charge — does not consume quota:**

- The automatic onboarding run does **not** count against the plan's `maxVolumeAnalyses` quota: skip the `volume_usage` insert for this internal run. It's a one-time welcome analysis; the user keeps their full monthly manual-analysis budget.

**Idempotency / safety:**

- The existing analyze logic already reuses saved keywords (LLM is only called for prompts without them), so a later manual *Analyze* click just refreshes volumes — no double LLM spend.
- If the brand already has volume rows (e.g. retriggered success URL), the run is a cheap refresh; guard with an "already analyzed recently" check if needed.
- DataForSEO/LLM errors are caught per prompt (as in the batch route) and logged; partial results are fine.

**Scope:**

- Cloud post-payment flow only. Self-host has no payment step (tracking starts at onboarding finish) — out of scope here; the `analyzePromptVolume` env checks already no-op when DataForSEO credentials are absent.
- The additional-brand wizard (Dashboard → Brands → New) has the same manual step — candidate for a follow-up using the same shared function.

## Acceptance criteria

- Completing checkout results in volume & competition data appearing on the Prompts page within a few minutes, with no manual action.
- `volume_usage` is not incremented by the automatic run; the manual *Analyze* button and its quota behave exactly as before.
- A DataForSEO or LLM outage during the auto-run never affects the checkout redirect or tracking kickoff.
- `cd server && yarn test` passes; the shared function is covered by the existing batch-route behavior.

BharadwajKanneveti:
## Summary

This PR automatically triggers prompt volume analysis after a successful Stripe checkout, so new users don't have to manually click **Analyze prompts** before seeing Volume and Competition data.

### Changes
- Extracted the shared prompt volume analysis logic into `server/src/lib/volume-analysis.js`.
- Updated the manual `/api/volumes/analyze-batch` endpoint to reuse the shared implementation.
- Added a new internal endpoint `POST /api/internal/analyze-volumes` secured with `CRON_SECRET` authentication.
- The internal endpoint starts the analysis asynchronously and immediately returns `202 Accepted`.
- Triggered the new internal endpoint from the Stripe success flow immediately after the existing tracking kickoff.
- Automatic onboarding analysis does **not** consume the user's monthly volume analysis quota; manual analysis behavior remains unchanged.

## Related issue

Closes #541

## Type of change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Checklist

### DIFF
--- a/server/src/lib/volume-analysis.js
+++ b/server/src/lib/volume-analysis.js
@@ -83,7 +83,10 @@ export async function fetchAndSaveVolumes(promptId, keywords, intent, locationCo
   return saved;
 }
 
-export async function analyzeBrandVolumes(brandId, { force = false } = {}) {
+export async function analyzeBrandVolumes(
+  brandId,
+  { promptIds, locationCode, languageCode, force = false, onlyMissing = false } = {},
+) {
   const { data: brand, error: brandError } = await supabaseAdmin
     .from('brands')
     .select('region, language')
@@ -97,9 +100,10 @@ export async function analyzeBrandVolumes(brandId, { force = false } = {}) {
     throw new Error(`Brand ${brandId} not found`);
   }
 
-  const resolvedLocationCode = brand.region ? regionToLocationCode(brand.region) : undefined;
-
-  const resolvedLanguageCode = brand.language ? languageToCode(brand.language) : undefined;
+  const resolvedLocationCode =
+    locationCode ?? (brand.region ? regionToLocationCode(brand.region) : undefined);
+  const resolvedLanguageCode =
+    languageCode ?? (brand.language ? languageToCode(brand.language) : undefined);
   const { data: promptSets, error: psError } = await supabaseAdmin
     .from('prompt_sets')
     .select('id')
@@ -114,11 +118,16 @@ export async function analyzeBrandVolumes(brandId, { force = false } = {}) {
   }
 
   const setIds = promptSets.map((ps) => ps.id);
+  let promptsQuery = supabaseAdmin.from('prompts').select('id, text');
 
-  const { data: prompts, error: pError } = await supabaseAdmin
-    .from('prompts')
-    .select('id, text')
-    .in('prompt_set_id', setIds);
+  if (promptIds?.length) {
+    promptsQuery = promptsQuery.in('id', promptIds);
+  } else {
+    promptsQuery = promptsQuery.in('prompt_set_id', setIds);
+  }
+
+  const { data: prompts, error: pError } = await promptsQuery;
+
   if (pError) {
     throw new Error(`Failed to fetch prompts: ${pError.message}`);
   }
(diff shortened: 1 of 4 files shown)


gkhngyk (changes requested):
Thanks — the shared `volume-analysis.js` extraction and the internal endpoint shape (CRON_SECRET + detached 202) are exactly right. Two things need fixing before this can land:

**1. `/analyze-batch` no longer respects the caller's selection.** The old endpoint analyzed the specific `prompts` array from the request body (the subset the user ticked in the Analyze dialog). The refactor resolves the brand from the first prompt and calls `analyzeBrandVolumes(brandId)`, which analyzes **every prompt in the brand**:

- Selecting 3 prompts on a 200-prompt brand now runs 200 serial LLM + DataForSEO calls inside one awaited request (cost + almost certain timeout).
- `volume_usage` gets recorded for all analyzed prompts, not the user's selection.
- The web client still sends `locationCode` / `languageCode` / `force` (`web/src/lib/actions/volumes.ts` → `analyzePromptVolumesBatch`), and the location/language overrides are now silently dropped.

Suggestion: give `analyzeBrandVolumes` optional `{ promptIds, locationCode, languageCode }` and pass the request's subset + overrides through from `/analyze-batch`, keeping its old contract intact.

**2. The auto-trigger isn't idempotent.** The Stripe success route runs on every visit with a `session_id` (its only guard is `payment_status === 'paid'`), so a refresh of the success page re-fires the analysis — cached intents are reused, but `fetchAndSaveVolumes` still re-hits DataForSEO for every prompt. Add an `onlyMissing: true` mode for the internal path that skips prompts which already have a `prompt_volumes` row: the first post-checkout call analyzes everything (all prompts are new), and any repeat call becomes a no-op.

Minor: the two import lines from `../lib/volume-analysis.js` in `routes/volumes.js` can be merged into one.

CI is green and the rest reads well — happy to approve once these two are in.

gkhngyk (approved):
Both requested changes are addressed — analyze-batch now scopes to the selected prompt IDs, and the post-checkout trigger is idempotent via onlyMissing. Thanks for the quick turnaround!

PR #583, opened 2026-08-03
fix(server): count run results from the DB when stamping tracking_runs

gkhngyk:
## Summary

**Urgent follow-up to #578.** This morning's cron ran on the new code but `tracking_runs` ended the day empty, so the insights 24h anchor never advanced past the backfill (the dashboard kept showing yesterday's run) and the Daily Pulse email disagreed with the product again.

Root cause: the zero-result guard read the worker's own `insertedCount` — but in **webhook mode the scraper results are inserted by the `/cloro/callback` handler**, not by the worker. The counter only sees the API-model phase, so on scraper-only brands every *successful* run ended with `insertedCount === 0` and the guard **deleted the just-completed run's ledger row**.

Fix:

- Stamp from a `prompt_results` count scoped to the run's `started_at` (insert-path-agnostic) — `result_count` now reflects the real run size.
- The delete only fires for genuinely zero-result runs; a count error leaves the row uncompleted (dashboard keeps the previous completed run; the next run clears the dangling row).
- Added a `tracking run stamped` info log so the next diagnosis is one grep away.

Today's missing run rows were repaired directly in the ledger (synthetic completed rows for the 20 brands that ran), so dashboards are already showing today's data — this PR makes tomorrow's 05:00 TR run stamp itself correctly.

## Related issue

Follow-up to #578 (no tracked issue).

## Type of change

- [x] `fix` — Bug fix

## How to test

1. Deploy the server, wait for (or trigger) a full run on a scraper-only brand.
2. `tracking_runs` gains a completed row whose `result_count` matches the run's `prompt_results` rows, and the server logs `tracking run stamped`.
3. The insights 24h view advances to the new run when it completes.

## Checklist

### DIFF
--- a/server/src/lib/pulse/metrics.js
+++ b/server/src/lib/pulse/metrics.js
@@ -298,6 +298,22 @@ async function degradedPlatforms(now) {
 }
 
 /**
+ * The last completed tracking runs' stamps, newest first (00044 ledger).
+ * Empty when the brand has never completed a full run.
+ */
+async function latestCompletedRunTimes(brandId, limit = 3) {
+  const { data, error } = await supabaseAdmin
+    .from('tracking_runs')
+    .select('completed_at')
+    .eq('brand_id', brandId)
+    .not('completed_at', 'is', null)
+    .order('completed_at', { ascending: false })
+    .limit(limit);
+  if (error) return [];
+  return (data ?? []).map((r) => new Date(r.completed_at));
+}
+
+/**
  * Compute the full pulse payload for a brand.
  *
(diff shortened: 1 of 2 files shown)


gkhngyk:
Added a second commit: the Daily Pulse's daily KPI windows now resolve from the tracking_runs ledger too — the same windows the Insights 24h view uses — so the email and the dashboard read identical ranges by construction. Wall-clock windows double-counted on the cron-time transition day (522 vs 278). Weekly pulses, detectors, and brands with no completed run keep clock-based windows.

PR #609, opened 2026-08-03
fix(web): show distinct error state with retry for failed suggestion …

### LINKED ISSUE #592: Prompts: failed suggestion loads are indistinguishable from "no suggestions"
### Problem

When the prompt suggestions card fails to load, the error is swallowed into `console.error` and no error state is set. Because `setLoaded(true)` is skipped on failure, the card renders the same "No suggestions right now" empty state as a genuinely empty result — telling the user to click Refresh when the real problem was a failed request.

### Evidence

- [`prompts/_suggestions-card.tsx#L75-L86`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/_suggestions-card.tsx#L75-L86) — `load()` catch block only logs
- The empty state that gets shown instead: [`#L214-L230`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/_suggestions-card.tsx#L214-L230)
- The pattern to copy is a few lines below in the same file: `handleRefresh` surfaces failures via `toast.error` ([`#L107-L119`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/_suggestions-card.tsx#L107-L119))

### Suggested fix

On catch, set an error flag and render a distinct "Couldn't load suggestions" state with a retry button (re-invoking `load()`), or at minimum fire the same `toast.error` that `handleRefresh` already uses.

### Acceptance criteria

- A failed load is visually distinguishable from "no suggestions available".
- The user has an explicit retry affordance after a failure.

d180:
## Summary

When the prompt suggestions card failed to load, the error was swallowed into `console.error` and the card silently rendered the same "No suggestions right now" empty state as a genuinely empty result — telling the user to click Refresh when the real problem was a failed request.

- Added an `error` state to `SuggestionsCard`, cleared on load start, brand change, and successful refresh
- `load()`'s catch block now sets the error and fires `toast.error`, matching the pattern already used by `handleRefresh`
- A new distinct "Couldn't load suggestions" state renders instead of the empty state when the load failed, with a Retry button that re-invokes `load()`

## Related issue

Closes #592

## Type of change

- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/_suggestions-card.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/_suggestions-card.tsx
@@ -15,6 +15,7 @@ import {
   Info,
   ChevronDown,
   ChevronRight,
+  AlertTriangle,
 } from 'lucide-react';
 import { toast } from 'sonner';
 import {
@@ -37,6 +38,7 @@ export function SuggestionsCard({ brandId, onAccepted }: Props) {
   const [suggestions, setSuggestions] = useState<PromptSuggestion[]>([]);
   const [loading, setLoading] = useState(false);
   const [loaded, setLoaded] = useState(false);
+  const [error, setError] = useState<string | null>(null);
   const [refreshing, setRefreshing] = useState(false);
   const [pendingId, setPendingId] = useState<string | null>(null);
   // Collapsed by default so the card is a one-line strip and the prompt table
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Clean error state — cleared in all the right places and the retry wires straight back into load(). Thanks!

PR #616, opened 2026-08-03
fix: filter zero-visit platforms from traffic breakdown

### LINKED ISSUE #606: AI Traffic: platforms with zero visits in the selected period inflate the Platforms KPI and table
### Problem

`getTrafficSummary` builds `platformBreakdown` from the **union** of the current and previous window platform maps, so a platform that only had traffic in the previous period is returned with `visits: 0`. Downstream this:

- inflates the "Platforms" KPI ("unique AI sources" counts platforms with zero visits in the selected period),
- adds `0`-visit rows with a `-100%` badge to the "Referrals by Platform" table,
- while the pie/bar chart on the same screen filters `visits > 0` — so the chart and the table/KPI disagree about how many platforms there are.

### Evidence

- Union + zero rows: [`traffic.ts#L203-L210`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/traffic.ts#L203-L210)
- KPI = `platformBreakdown.length`: [`traffic/page.tsx#L684-L689`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/traffic/page.tsx#L684-L689)
- Chart filters zero rows: [`traffic/_charts.tsx#L252-L253`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/traffic/_charts.tsx#L252-L253)

### Suggested fix

The union is needed to compute deltas, so keep it in the action — but drop `visits === 0` entries from the returned `platformBreakdown` (or filter at the two render sites). Decide the product behavior explicitly: a platform present yesterday but absent today probably shouldn't count as a current "unique AI source" or occupy a table row at `0 (-100%)`.

### Acceptance criteria

- The Platforms KPI counts only platforms with ≥ 1 visit in the selected period.
- The platform table no longer lists `0`-visit rows.
- The KPI, table, and chart agree on the platform count.

Jesulac:
Closes #606. Platforms that only existed in the previous period were appearing with visits=0 in the current period's platformBreakdown, inflating the Platforms KPI and showing 0-visit rows in the table. The chart already filtered these out, so this adds a matching filter to the data action to keep KPI, table, and chart consistent. All tests pass.

### DIFF
--- a/web/src/lib/actions/traffic.ts
+++ b/web/src/lib/actions/traffic.ts
@@ -207,6 +207,7 @@ export async function getTrafficSummary(
       visits: platformMap.get(platform) ?? 0,
       visitsPrev: platformMapPrev.get(platform) ?? 0,
     }))
+    .filter((p) => p.visits > 0)
     .sort((a, b) => b.visits - a.visits);
 
   const topPages = Array.from(pageMap.entries())


gkhngyk (approved):
Exactly the right one-line fix — keeps the union for deltas and drops the zero rows from the response. Thanks, and welcome!

PR #617, opened 2026-08-03
fix(web): count topic prompts by topic_id instead of category name

### LINKED ISSUE #594: Renaming a topic zeroes its prompt count on the Manage Topics page
### Problem

Renaming a topic on the Manage Topics page makes its prompt count drop to 0. `updateTopic` renames only the `topics` row, but `getPromptCountByTopic` counts prompts by the legacy `prompts.category` **string** — which still holds the old name after a rename. The Topics leaderboard, which counts by `topic_id`, keeps showing the correct number, so the two surfaces disagree.

### Reproduction

1. Create a topic with prompts assigned to it — Manage Topics shows the correct count.
2. Rename the topic.
3. The count next to it drops to 0; the Topics leaderboard still shows the real count.

### Evidence

- [`topic.ts#L84-L96`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic.ts#L84-L96) — `updateTopic` updates `{ name }` only
- [`topic.ts#L98-L115`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic.ts#L98-L115) — `getPromptCountByTopic` filters `.eq('category', topicName)`
- Caller passes the (new) name: [`brands/[id]/topics/page.tsx#L57`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/brands/%5Bid%5D/topics/page.tsx#L57)
- The correct counting path for contrast, keyed on `topic_id`: [`topic.ts#L288`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic.ts#L288)
- `category` and `topic_id` are written together on prompt save/update: [`prompt.ts#L279`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/prompt.ts#L279), [`#L326`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/prompt.ts#L326)

### Suggested fix

Either count by `topic_id` in `getPromptCountByTopic` (preferred — change the parameter to the topic id and the filter to `.eq('topic_id', topicId)`), or have `updateTopic` also sync `prompts.category` for rows pointing at the renamed topic. The first is a smaller, safer change.

### Acceptance criteria

- After renaming a topic, its prompt count on Manage Topics is unchanged.
- Manage Topics and the Topics leaderboard report the same count for the same topic.

d180:
## Summary

- Renaming a topic on Manage Topics zeroed its prompt count: `getPromptCountByTopic` matched prompts by the legacy `prompts.category` text field, which still held the topic's old name after a rename.
- The Topics leaderboard, which counts by `topic_id`, kept showing the correct number — so the two surfaces disagreed.
- Switches `getPromptCountByTopic` to filter by `topic_id` instead of `category` name (the smaller, safer of the two fixes suggested in the issue), and updates its caller in the Manage Topics page to pass the topic's id instead of its name.

## Related issue

Closes #594

## Type of change

- [x] `fix` — Bug fix

## Checklist

- [x] Branch follows the naming convention (`fix/`)

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/brands/[id]/topics/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/brands/[id]/topics/page.tsx
@@ -54,7 +54,7 @@ export default function BrandTopicsPage({ params }: PageProps) {
         const counts: Record<string, number> = {};
         await Promise.all(
           data.map(async (t) => {
-            counts[t.id] = await getPromptCountByTopic(brandId, t.name);
+            counts[t.id] = await getPromptCountByTopic(brandId, t.id);
           }),
         );
         if (isCancelled?.()) return;
--- a/web/src/lib/actions/topic.ts
+++ b/web/src/lib/actions/topic.ts
@@ -95,7 +95,7 @@ export async function updateTopic(topicId: string, name: string): Promise<Topic>
   return mapTopicRow(data as Record<string, unknown>);
 }
 
-export async function getPromptCountByTopic(brandId: string, topicName: string): Promise<number> {
+export async function getPromptCountByTopic(brandId: string, topicId: string): Promise<number> {
   const supabase = await createClient();
 
   const { data: sets } = await supabase.from('prompt_sets').select('id').eq('brand_id', brandId);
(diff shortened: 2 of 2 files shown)


gkhngyk (approved):
The preferred fix from the issue, applied cleanly at both the action and its only caller. Thanks!

PR #622, opened 2026-08-03
fix(web): populate the Citations Region filter from observed data

### LINKED ISSUE #598: Citations: Region filter dropdown is permanently empty
### Problem

The Region filter dropdown on the Citations page is permanently empty — it only ever shows "All Regions". `loadData` creates `const regions = new Set<string>()` and never adds anything to it, so the `if (regions.size > 0)` guard is always false and `availableRegions` stays `[]` forever. The select still renders, offering a filter that can never be used.

The filter *side* works: picking a region (if one existed) would be applied server-side. Only the option discovery is dead.

### Evidence

- [`citations/page.tsx#L860-L872`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L860-L872) — the never-populated `Set` and the dead `regions.size > 0` branch
- Always-rendered select fed by `availableRegions`: [`filter-bar.tsx#L342-L360`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/citations/filter-bar.tsx#L342-L360)
- The overview payload has no region data to read from: [`citations.ts#L67-L78`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts#L67-L78) (`CitationsOverview` type) — `region` is selected from `prompt_results` in the queries but never aggregated into the response

### Suggested fix

Two acceptable scopes — the issue can be closed with either:

1. **Small:** hide the Region select when `availableRegions` is empty (and delete the dead `Set`), so no unusable control is shown.
2. **Complete:** have `getCitationsOverview` return the distinct regions it already reads from `prompt_results`, populate `availableRegions` from that, and keep the select.

### Acceptance criteria

- Either the Region select is hidden when there are no options, or it lists the regions actually observed in the data.
- No dead `regions` code path remains.

d180:
## Summary

- The Region filter dropdown on the Citations page was permanently empty — `loadData` created a `Set<string>` for regions and never added anything to it, so the dropdown only ever showed "All Regions".
- `getCitationsOverview` already selects `region` from `prompt_results` while scanning the filtered window, but discarded it before it reached the client.
- `getCitationsOverview` now collects the distinct, non-null regions it observes and returns them as `availableRegions` on `CitationsOverview`.
- The page's `loadData` now populates `availableRegions` state from that field instead of the dead local `Set`, using the same accumulate-and-sort merge already used for platform options.

## Related issue

Closes #598

## Type of change

- [x] `fix` — Bug fix

## Checklist

- [x] Branch follows the naming convention (`fix/`)

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
@@ -857,7 +857,6 @@ export default function CitationsPage() {
 
       // Surface filter options from the observed models/regions.
       const platformOptions = buildPlatformOptions(overview.rows);
-      const regions = new Set<string>();
       setAvailablePlatforms((prev) =>
         Array.from(
           new Map(
@@ -865,9 +864,11 @@ export default function CitationsPage() {
           ).values(),
         ).sort((a, b) => a.label.localeCompare(b.label) || a.value.localeCompare(b.value)),
       );
-      if (regions.size > 0) {
+      if (overview.availableRegions.length > 0) {
         setAvailableRegions((prev) =>
-          Array.from(new Set([...prev, ...regions])).sort((a, b) => a.localeCompare(b)),
+          Array.from(new Set([...prev, ...overview.availableRegions])).sort((a, b) =>
+            a.localeCompare(b),
+          ),
         );
       }
     } catch {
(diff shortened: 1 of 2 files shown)


gkhngyk (approved):
Nice — piggybacking on the existing scan keeps this zero-cost, and the accumulate-and-merge keeps options stable while a region filter is active. Thanks!

PR #618, opened 2026-08-03
fix(web): show distinct error states with retry across Topics loaders

d180:
## Summary

- All three Topics data loaders swallowed failures into `console.error`, so a failed request rendered as a normal empty state — "No topics defined yet" or "Topic not found" — with no way to tell an outage from a genuinely empty account, and no retry.
- Adds an `error` state to each loader and a distinct error UI with a retry action:
  - `topics/page.tsx` — leaderboard card now shows a "Failed to load topics" error with retry instead of the empty state (uses the page's existing i18n).
  - `topics/[id]/page.tsx` — a failed load now shows a "Couldn't load this topic" card with retry; "Topic not found" is now reserved for a successful fetch that genuinely returns nothing.
  - `topics/_suggestions-card.tsx` — mirrors the pattern already shipped for the Prompts suggestions card (#previous fix), same error state + retry button.

## Related issue

Closes #595

## Type of change

- [x] `fix` — Bug fix

## Checklist

- [x] Branch follows the naming convention (`fix/`)
- [x] Commits follow Conventional Commits
- [x] `yarn lint` passes (run from `web/`)
- [x] `yarn typecheck` passes (run from `web/`)
- [x] `yarn format:check` passes (run from `web/`)
- [x] Changes are focused — one concern per PR

PR #627, opened 2026-08-03
fix: handle prompt detail fetch errors

### LINKED ISSUE #591: Prompt detail page hangs on the loading skeleton forever when the data fetch fails
### Problem

The prompt detail page awaits `getPromptDetail` with no `try/catch`. If the server action rejects (network hiccup, transient server error), the promise rejection skips straight past `setLoading(false)`, leaving the page on its skeleton forever — no error message, no retry, and the date-filter changes that re-run the effect are the user's only accidental way out.

### Evidence

- [`prompts/[id]/page.tsx#L629-L649`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/%5Bid%5D/page.tsx#L629-L649) — `load()` awaits `getPromptDetail` unguarded; `setLoading(false)` is the last line and is never reached on rejection
- The sibling list page handles the same failure correctly with `try/catch/finally` + `toast.error`: [`prompts/page.tsx#L946-L967`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/page.tsx#L946-L967)
- The workflow fetch in the same detail file already guards its own call ([`#L614`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/%5Bid%5D/page.tsx#L614)), so the gap is one function.

### Suggested fix

Wrap the body of `load()` in `try/catch/finally`: move `setLoading(false)` into `finally`, and on catch surface a `toast.error` (matching the list page) — an inline error state with a retry button is a welcome extra.

### Acceptance criteria

- A rejected `getPromptDetail` leaves the page in a non-loading state with a visible error, not a permanent skeleton.
- The `cancelled` guard still prevents state updates after unmount.

BharadwajKanneveti:
## Summary

Adds error handling to the prompt detail page to prevent it from getting stuck in a loading state when `getPromptDetail` fails.

### Changes
- Wrapped the prompt detail fetch in a `try/catch/finally` block.
- Moved `setLoading(false)` into `finally` so the loading state is always cleared, even if the request fails.
- Added `toast.error` to surface fetch failures to the user.
- Preserved the existing cancellation guard to avoid state updates after the component unmounts.

This ensures transient fetch failures no longer leave the page displaying a permanent loading skeleton.

## Related issue

Closes #591 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

```

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/[id]/page.tsx
@@ -628,21 +628,41 @@ export default function PromptDetailPage() {
 
   useEffect(() => {
     let cancelled = false;
+
     async function load() {
       setLoading(true);
-      const { dateFrom, dateTo } = getDateRange(datePreset, { from: customFrom, to: customTo });
-      const detail = await getPromptDetail(promptId, { dateFrom, dateTo });
-      if (cancelled) return;
-      if (!detail) {
-        setNotFound(true);
-      } else {
-        setData(detail);
-        const groups = groupByPlatform(detail.results);
-        setExpanded(new Set(groups.length > 0 ? [groups[0].key] : []));
+
+      try {
+        const { dateFrom, dateTo } = getDateRange(datePreset, {
+          from: customFrom,
+          to: customTo,
+        });
+
+        const detail = await getPromptDetail(promptId, { dateFrom, dateTo });
+
+        if (cancelled) return;
+
+        if (!detail) {
+          setNotFound(true);
+        } else {
+          setData(detail);
+          const groups = groupByPlatform(detail.results);
+          setExpanded(new Set(groups.length > 0 ? [groups[0].key] : []));
+        }
+      } catch (err) {
+        if (cancelled) return;
+
+        console.error('Failed to load prompt detail:', err);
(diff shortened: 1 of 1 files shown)


gkhngyk (changes requested):
The try/catch/finally structure is right and the endless skeleton is fixed. One small change before merge: after a failed fetch, `data` stays null so the page falls into the existing `notFound || !data` branch and shows "Prompt not found — may have been deleted", which is misleading for a transient network error.

Suggestion: add a `loadFailed` state set in the catch, and render a distinct "Couldn't load this prompt" message (a Retry button re-invoking the load would be a nice touch) before the not-found branch — keeping "Prompt not found" for a successful fetch that returns nothing. #618 established this pattern on the Topics pages if you want a reference.

BharadwajKanneveti:
Hi @gkhngyk I have made the changes please check once.

gkhngyk (approved):
Exactly what was asked — the distinct error card with retry before the not-found branch, and "Prompt not found" now reserved for a successful-but-empty fetch. Thanks!

PR #632, opened 2026-08-03
fix: handle audit polling timeout state

### LINKED ISSUE #612: Site Audit: detail page hangs on the spinner forever when polling times out — no retry, actions hidden
### Problem

The audit detail page polls `getAudit` every 3s with a 120-second deadline. If the audit is still `running` when the deadline expires, the loop exits silently: `audit.status` stays `'running'`, the page keeps showing the spinner card ("~30s") indefinitely, and — because the header's re-run and delete buttons are hidden precisely while `status === 'running'` — the user has no affordance left except a manual browser reload. The stage label also freezes on the last stage after ~16s, reinforcing the "stuck" impression.

### Evidence

- Deadline exits with no state change: [`audit/[id]/page.tsx#L47-L56`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/%5Bid%5D/page.tsx#L47-L56)
- Actions hidden while running: [`#L111`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/%5Bid%5D/page.tsx#L111) — `{audit && audit.status !== 'running' && (`
- The spinner branch with the hardcoded `~30s`: [`#L152-L163`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/%5Bid%5D/page.tsx#L152-L163)
- Stage label caps at 2: [`#L76-L80`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/%5Bid%5D/page.tsx#L76-L80)

### Suggested fix

Track a `timedOut` flag when the deadline expires while still running. Render a distinct card ("This audit is taking longer than expected") with a **"Check again"** button that restarts the poll loop, and stop hiding the delete action in that state so the user can clean up a genuinely stuck audit. Add the new strings to the `audit` namespace in `messages/en.json`.

### Acceptance criteria

- After the polling deadline, the user sees an explicit "taking longer than expected" state with a working "Check again" action — never an endless spinner.
- The audit becomes viewable if a later check finds it completed.

BharadwajKanneveti:
## Summary

Fix audit detail page polling timeout handling. Previously, when an audit remained in `running` state after the 120-second polling deadline, the page would silently stop polling and keep showing an infinite spinner with no available actions.

This change:
- Tracks a `timedOut` state when polling reaches the deadline while the audit is still running.
- Shows a dedicated "taking longer than expected" state instead of an endless spinner.
- Adds a "Check again" action to restart polling.
- Restores header actions (re-run/delete) when an audit times out.
- Adds translations for the new timeout state messages.

## Related issue

Closes #612 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/messages/en.json
+++ b/web/messages/en.json
@@ -73,6 +73,8 @@
     "noBrand": "Select a brand first.",
     "urlRequired": "Enter a URL to audit.",
     "failed": "Audit failed. Please try again.",
+    "takingLongerThanExpected": "This audit is taking longer than expected.",
+    "checkAgain": "Check again",
     "grade": {
       "good": "Good",
       "needsWork": "Needs work",
--- a/web/src/app/[locale]/(dashboard)/dashboard/audit/[id]/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/audit/[id]/page.tsx
@@ -32,46 +32,47 @@ export default function AuditDetailPage() {
   const [loading, setLoading] = useState(true);
   const [stage, setStage] = useState(0);
   const [timedOut, setTimedOut] = useState(false);
-  const cancelledRef = useRef(false);
+  const genRef = useRef(0);
   const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   const loadAndPoll = useCallback(async () => {
+    const gen = ++genRef.current;
+    const cancelled = () => genRef.current !== gen;
     try {
-      cancelledRef.current = false;
       setTimedOut(false);
 
       let result = await getAudit(id);
 
-      if (cancelledRef.current) return;
+      if (cancelled()) return;
 
       setAudit(result);
       setLoading(false);
 
       const deadline = Date.now() + 120_000;
 
-      while (!cancelledRef.current && result.status === 'running' && Date.now() < deadline) {
+      while (!cancelled() && result.status === 'running' && Date.now() < deadline) {
         await new Promise((resolve) => {
           timerRef.current = setTimeout(resolve, 3000);
         });
 
-        if (cancelledRef.current) return;
+        if (cancelled()) return;
 
         result = await getAudit(id);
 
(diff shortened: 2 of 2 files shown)


gkhngyk (changes requested):
The timeout handling itself is right — the dedicated state, Check again with a fresh deadline, and unhiding the header actions all match #612. One correctness issue in the refactor before merge:

The original effect used a per-invocation `let cancelled`; this PR switches to a shared `cancelledRef` that `loadAndPoll` resets to `false` on every call. That opens a race when `id` changes without an unmount (navigating /audit/A → /audit/B re-renders with new params):

1. A's loop is asleep in its 3s timeout; cleanup sets `cancelledRef = true`.
2. The new effect immediately calls `loadAndPoll`, which resets `cancelledRef = false`.
3. A's loop wakes, sees `false`, and keeps polling the old id — `setAudit` then writes A's results onto B's page.

Suggested fix — a generation counter instead of a shared boolean:

```ts
const genRef = useRef(0);

const loadAndPoll = useCallback(async () => {
  const gen = ++genRef.current;          // this call's token
  const cancelled = () => genRef.current !== gen;
  // ... use cancelled() everywhere cancelledRef.current is checked
}, [id, tFailed]);

// cleanup: genRef.current++; clearTimeout(...)
```

Each call owns its token, so starting a new poll (Check again or an id change) automatically invalidates every older loop with no shared flag to reset. Happy to re-review once that's in.

gkhngyk (approved):
The generation-counter fix is exactly right — cleanup only ever advances the counter, so stale loops can never resurrect. Thanks for the quick turnaround!

PR #640, opened 2026-08-04
fix(server): generate content opportunities and briefs in the brand's…

### LINKED ISSUE #638: Content Opportunities and briefs are always generated in English regardless of the brand's language
### Problem

Content Opportunities and their briefs are always generated in English, even for brands whose language is not English (e.g. a Turkish-market brand with `language: tr` gets English titles, descriptions, and briefs).

The cause is that neither generation path ever tells the model the brand's language:

1. **Opportunity generation** — the brand query selects only `id, name, description, industry`; the `brands.language` / `brands.region` columns (which exist and are used by volume analysis) are never read. The system prompt contains no output-language instruction, so the model answers in the instruction's language: English.
2. **Brief generation** — same gap: the brand select has no `language`, and the brief system prompt has no language rule. So even a manually translated opportunity would still get an English brief.

`grep -c language server/src/routes/content.js` returns 0 — the language is never forwarded anywhere in the content pipeline.

### Evidence

- Opportunity brand select (no `language`): [`opportunity-generator.js#L65-L69`](https://github.com/ansvisor/ansvisor/blob/main/server/src/lib/opportunity-generator.js#L65-L69)
- English-only system prompt, no output-language rule: [`opportunity-generator.js#L35-L42`](https://github.com/ansvisor/ansvisor/blob/main/server/src/lib/opportunity-generator.js#L35-L42)
- User prompt without language: [`opportunity-generator.js#L156-L164`](https://github.com/ansvisor/ansvisor/blob/main/server/src/lib/opportunity-generator.js#L156-L164)
- Brief brand select (no `language`): [`content.js#L602-L606`](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/content.js#L602-L606)
- Brief system prompt, no language rule: [`content.js#L555`](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/content.js#L555)
- The columns exist with defaults `region: 'US'`, `language: 'en'` — see `brands` in [`supabase/schema.sql`](https://github.com/ansvisor/ansvisor/blob/main/supabase/schema.sql)

### Suggested fix

Symmetric ~10-line changes in the two files:

1. Add `language` to both brand selects.
2. Append an explicit instruction to both user prompts (or system prompts), e.g.:
   `Output language: ${brand.language || 'en'} — write every title, description, and brief field in this language. Keep proper nouns and metric names as-is.`

No migration and no re-analysis needed — the next generation cycle picks up the language. Existing English rows stay untouched.

### Acceptance criteria

- A brand with `language: tr` gets Turkish opportunity titles/descriptions and Turkish briefs on the next generation.
- Brands with `language: en` (the default) see no change.
- Proper nouns (brand/competitor names) are not translated.

d180:
## Summary

- Content Opportunities and their AI-generated briefs were always produced in English, even for brands with a non-English `brands.language` (e.g. `tr`), because neither generation path read the `language` column or told the model what language to write in.
- Add `language` to the brand `.select(...)` in both `server/src/lib/opportunity-generator.js` (opportunity generation) and `server/src/routes/content.js` (brief generation).
- Pass `getLanguageName(brand.language)` into each prompt with an `IMPORTANT: Write ... in ${langName}.` instruction — matching the same convention already used in `content-worker.js`, `topics.js`, `competitors.js`, and `prompts.js`.
- No migration or re-analysis needed: `language` already defaults to `'en'` on `brands`, so English brands see no behavior change; the fix applies on the next generation cycle.

## Related issue

Closes #638

## Type of change

- [x] `fix` — Bug fix

## Checklist

- [x] `npm run lint` passes (run from `server/`; this PR only touches `server/`, no `web/` changes)
- [x] `npm run format:check` passes (run from `server/`)
- [x] `npm run test` passes (run from `server/`, 147/147 tests)

### DIFF
--- a/server/src/lib/opportunity-generator.js
+++ b/server/src/lib/opportunity-generator.js
@@ -6,6 +6,7 @@
 import { generateObject } from 'ai';
 import { z } from 'zod';
 import { resolveModel } from './ai-provider.js';
+import { getLanguageName } from './languages.js';
 import supabaseAdmin from '../config/supabase.js';
 import { logger } from './logger.js';
 
@@ -64,10 +65,12 @@ export async function generateContentOpportunities(brandId) {
 
   const { data: brand } = await supabaseAdmin
     .from('brands')
-    .select('id, name, description, industry')
+    .select('id, name, description, industry, language')
     .eq('id', brandId)
     .single();
   if (!brand) return;
+
+  const langName = getLanguageName(brand.language);
 
   const { data: domains } = await supabaseAdmin
     .from('brand_domains')
@@ -161,7 +164,9 @@ Competitors: ${(compRes.data || []).map((c) => c.name).join(', ') || 'None'}
 Prompt Data:
 ${scored.map((p, i) => `[${i}] "${p.text}" | Intent: ${p.intent} | AI Vol: ${p.estAiVolume}/mo | Vis: ${p.avgVisibility}% | Gap: ${p.competitorGap}%`).join('\n')}
 
-Generate actionable content opportunities.`;
+Generate actionable content opportunities.
+
+IMPORTANT: Write every opportunity title and description in ${langName}.`;
 
   const { object } = await generateObject({
     model: resolveModel(),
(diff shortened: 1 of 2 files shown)


gkhngyk (approved):
Exactly the fix from the issue, and good call reusing the existing languages.js map. One optional follow-up: the issue's acceptance notes say proper nouns (brand/competitor names) shouldn't be translated — appending "keep proper nouns and metric names as-is" to the language instruction would pin that down. Fine as-is. Thanks!

Issue #427, opened 2026-07-15
Query fan-out: long hang then production-masked server error surfaced raw in the UI


gkhngyk:
## Bug

On the prompts page's **Query Fan-out** tab (and its two views, frequency / by-prompt), the tab can hang for a long time and then show this raw message in a toast — reproduced in production, also when promoting a sub-query with the Track (+) button:

> An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance…

Both views die together because they share one `load()` path.

## What's wrong (multiple compounding defects)

1. **Raw digest message reaches the user.** `web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx` catches the failed server action and does `toast.error(err.message)` — in production that message is Next's masked digest text, meaningless to users. It should map to a friendly "couldn't load fan-out" string (and the digest belongs in server logs, not the toast).

2. **`getQueryFanout` scans brand-wide and truncates silently.** https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/fanout.ts fetches every `prompt_results` row for the brand in the window with the full `search_queries` jsonb and aggregates in JS. Two problems:
   - For busy brands this is slow and can push the server action toward the function time limit — matching the observed "long hang then error".
   - Supabase/PostgREST caps un-paginated selects at 1000 rows **silently**, so the aggregation is quietly incomplete on exactly the brands where fan-out matters most. Fix: paginate with `.range()` (or move the aggregation into an RPC like the insights surfaces from #114).

3. **`classifyFanoutIntents` fetch has no timeout.** The server action fetches `${API_BASE_URL}/api/prompts/fanout-intents`, which classifies up to 200 cache-miss queries via LLM at concurrency 6 — minutes on a cold cache. If the express server is slow or hung, the action hangs until the platform kills it. The initial load races it against a 1.5s timeout (good), but the action itself should still carry an `AbortController` timeout (~20s) so it can never pin a function.

4. **Track (+) path**: `trackFanoutQuery` → `addPromptToSet` chains 7+ sequential DB round-trips (set lookup, brand context, org prompt count, plan fetch, topic resolve, shopping-mode read, insert). Worth a look while in here, since the masked error also reproduces on Track.

## Diagnosis needed

[...]

PR #651, opened 2026-08-05
feat: show fan-out coverage next to sub-query count in the by-prompt view

### LINKED ISSUE #543: Fan-out by-prompt view: show run count / fan-out coverage next to sub-query count
## Problem

In **Prompts → Fan-out → By prompt**, each row shows only the number of distinct sub-queries. Users can't tell how that relates to tracking volume: a prompt with 500 tracking runs and 1 observed fan-out means the answer engines are essentially *not* searching for that prompt — a very different story from 10 runs / 8 fan-outs. Without the denominator, the signal is invisible.

## Proposal

Add fan-out coverage to the by-prompt table: alongside the sub-query count, show **runs with fan-out / total runs** in the window (e.g. `12 / 500 · 2%`).

**Data — zero extra queries:** `fetchFanoutRows` in [web/src/lib/actions/fanout.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/fanout.ts) already fetches *all* `prompt_results` rows in the window (there is deliberately no server-side non-empty filter on `search_queries`; empty rows are skipped in JS). Extend the same aggregation pass to also count, per prompt:

- `totalRuns` — all fetched rows for the prompt,
- `runsWithFanout` — rows where at least one valid sub-query was observed.

Expose both on the by-prompt payload and render in `_fanout-tab.tsx` (by-prompt view), e.g. a muted `12/500 (2%)` cell or badge next to the existing sub-query count.

**Notes:**

- Counts are scoped to the same fetched window (`days` + `FANOUT_MAX_ROWS` cap) as the sub-query data, so numerator and denominator always describe the same slice.
- Low coverage is expected for engines that rarely search (e.g. plain ChatGPT web answers) — a per-engine coverage breakdown could be a v2; v1 keeps the single ratio.
- Tooltip copy suggestion: "Of N tracked answers in this window, X triggered live searches."

## Acceptance criteria

- By-prompt rows show sub-query count *and* runs-with-fanout / total-runs (+ percentage) for the selected window.
- Numbers are consistent with the frequency view's data (same fetch, same window).
- A prompt with many runs and zero fan-outs clearly reads as 0/N (0%) rather than just an absent row.

arambu1a:
Closes #543

## Problem

In Prompts → Fan-out → By prompt, each row showed only the number of distinct sub-queries, with no denominator. "8 sub-queries" read identically whether it came from 10 tracked answers or 500 — and a prompt the engines never search for was invisible entirely, because the table was built by inverting sub-queries → prompts, so a prompt with zero fan-out produced no row at all.

## What changed

`getQueryFanout` now returns `promptCoverage: { id, text, totalRuns, runsWithFanout }[]`, accumulated in the **same single pass over `prompt_results`** that already builds the sub-query list. Numerator and denominator therefore describe the same window by construction — the two views cannot disagree, and there are no extra queries.

The By-prompt table is seeded from `promptCoverage` before the sub-query inversion, so every prompt with a run in the window gets a row. A new "Answers with fan-out" column renders `12 / 500 · 2%`, with a tooltip: *"Of N tracked answers in this window, X triggered live searches."* Zero-fan-out rows are inert (no expand chevron) rather than opening an empty panel.

Two pre-existing silent-truncation bugs surfaced while wiring this up and are fixed here:

- the tracked-prompt lookup was un-paginated, so PostgREST silently capped it at 1000 rows — brands with large prompt sets were getting an incomplete "is this tracked?" map. Now paged, alongside the existing result paging (#427/#428).
- `fetchFanoutRows` hit its row ceiling and returned with no signal. It now reports `truncated`, and the tab says counts are a lower bound instead of presenting a confident wrong ratio.

Net effect on query count is a reduction: one roster read now serves both prompt-text resolution and the tracked-prompt map (3 reads → 2 in the common case), with the residual `.in('id', …)` chunk skipped entirely when the roster covers every prompt.

Also: the expand chevron is now a real focusable `<button>` with `aria-expanded`, and the By-prompt list has its own pager since it covers every tracked prompt rather than only those with fan-out.

Screenshot 2026-08-05 at 1 41 16 PM

## Limitations

`totalRuns` spans all platforms. A prompt tracked on ChatGPT + Perplexity where Perplexity fans out every time reads as ~50%, which reflects engine mix more than search behaviour. Per-engine coverage breakdown is the v2 noted in the issue.

## Testing

14 tests in `fanout.test.ts` over a mocked Supabase client: `0 / N` for a prompt that ran but never searched, partial coverage, per-answer dedup, blank/non-string sub-queries counting toward the denominator only, `runsWithFanout ≤ totalRuns`, paused prompts included in coverage but not marked "Tracked ✓", the one-roster-read contract, fail-loud on a residual chunk error, and both truncation branches.

### DIFF
--- a/supabase/seed.sql
+++ b/supabase/seed.sql
@@ -177,6 +177,15 @@ INSERT INTO public.prompts (
    'Manual vs automatic espresso machine for home use',
    ARRAY['chatgpt-web','claude','copilot-web','gemini-web','grok-web'],
    ARRAY['US','TR','DE'],
+   ARRAY['claude-sonnet-4-6'], true),
+  -- Tracked ONLY on engines that never fan out, and factual enough that they
+  -- answer it from parametric knowledge. It is the Query Fan-out tab's
+  -- zero-coverage case: many tracked answers, not one live search. See the
+  -- "Zero fan-out coverage" block at the end of this file.
+  ('88888888-8888-8888-8888-888888888807', '66666666-6666-6666-6666-666666666601', '77777777-7777-7777-7777-777777777702',
+   'How much caffeine is in a double espresso?',
+   ARRAY['chatgpt-web','claude'],
+   ARRAY['US','GB','DE','FR','TR'],
    ARRAY['claude-sonnet-4-6'], true)
 ON CONFLICT (id) DO NOTHING;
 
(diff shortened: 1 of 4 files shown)


gkhngyk (approved):
Excellent work — this is exactly what #543 asked for, and then some. The coverage counters accumulating in the same pass that builds the sub-query list means the two views can never disagree, seeding the table from coverage makes the 0/N rows finally visible, and catching the two silent-truncation bugs (un-paginated tracked-prompt map, ceiling exit with no signal) along the way is a real bonus. The test suite pins all the right contracts. Thanks!

Issue #656, opened 2026-08-07
Citations: keep content visible during refetch instead of swapping the whole page to a skeleton


gkhngyk:
### Problem

The Citations page has a single `isLoading` flag, and while it is true the entire content — KPI cards, tables, tab counts — is replaced by `CitationsSkeleton`. That happens on **every** load, not just the first one: changing the date range or any filter blanks the page and re-renders it from grey boxes. On slow loads this reads as "the page reset" rather than "your data is updating", and all context (scroll position, which numbers were on screen) is lost.

The Insights page already solves this correctly and the pattern just needs to be copied.

### Evidence

- Skeleton swap on every load: [web/src/app/%5Blocale%5D/(dashboard)/dashboard/citations/page.tsx#L1106-L1108](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L1106-L1108)
- `loadData` sets the same `isLoading` for initial load and refetches alike: [page.tsx#L878-L916](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L878-L916)
- The pattern to copy — Insights distinguishes initial load from refetch: [web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx#L1349](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/page.tsx#L1349) (`const isRefetching = isLoading && summary !== null;`) and [insights/page.tsx#L1482-L1489](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/page.tsx#L1482-L1489) — stale content stays mounted, dimmed to 60% opacity, with a small floating spinner.

### Proposal

- Derive `isRefetching = isLoading && data !== null` on the Citations page.
- Show `CitationsSkeleton` only when `isLoading && data === null` (true first load).
- During a refetch, keep the existing content rendered, dim it (`opacity-60`), and show the same floating `Loader2` spinner Insights uses.
[...]

Issue #614, opened 2026-08-03
Site Audit: deleting from the recent-audits list has no confirmation (detail page has one)


gkhngyk:
### Problem

Deleting an audit from the hub's recent-audits list fires immediately on the trash-icon click, with an optimistic removal and no confirmation. The detail page guards the very same action behind a confirm `Dialog`, and the `deleteConfirm` copy already exists in `messages/en.json` — so the list is both a data-loss hazard and inconsistent with its sibling.

### Evidence

- Unconfirmed delete in the list: [`audit/page.tsx#L333-L341`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/page.tsx#L333-L341) — `onClick={() => handleDelete(h.id)}`
- Optimistic handler: [`#L138-L147`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/page.tsx#L138-L147)
- The confirm pattern already in the feature: [`audit/[id]/page.tsx#L122-L147`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/%5Bid%5D/page.tsx#L122-L147) — Dialog with `t('deleteConfirm')`

### Suggested fix

Wrap the list's trash button in the same `Dialog` used on the detail page (title `deleteAudit`, description `deleteConfirm`, destructive confirm). No new copy is needed.

**Small adjacent polish, same file:** for non-completed audits the list renders the raw status string in the score slot ([`#L330`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/audit/page.tsx#L330) — `: h.status`), so users see lowercase `running` / `failed` where a score belongs. Render a small status badge (or translated label) instead.

### Acceptance criteria

- Deleting from the list requires the same confirmation as the detail page; cancelling keeps the audit.
- Non-completed rows show a presentable status label instead of the raw enum string.

Adarshhic:
Hi @gkhngyk, I'd like to work on this one — could you assign it to me?

I've gone through the code and confirmed the issue:
- `audit/page.tsx#L333-L341` fires `handleDelete(h.id)` directly on click, with the optimistic removal in `handleDelete` (L138-147) meaning the row disappears before any confirmation.
[...]

PR #676, opened 2026-08-07
fix: keep filter bar visible when no results match active filters

### LINKED ISSUE #660: Content: zero-result filters hide the filter bar, leaving no way to clear them
### Problem

The Content list decides between empty state and content with `opportunities.length === 0` — but `opportunities` is the **server-filtered** page. Select a filter combination that matches nothing (e.g. Status = dismissed on a brand with no dismissed rows) and the whole content branch — including the Card that holds the filter selects — unmounts. The filter value lives in React state, so the user is stuck: no controls on screen to undo their own filter without reloading the page. Worse, the empty state shown is the "No content opportunities yet" + Generate card, which is simply wrong — data exists, the filter just matched nothing.

### Evidence

- Empty-state gate: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L426](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L426)
- Filter bar rendered only in the else branch: [page.tsx#L468-L534](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L468-L534)
- Filters are sent server-side: [page.tsx#L172-L182](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L172-L182)
- An in-table "No opportunities match your filters." message already exists but is unreachable in this state: [page.tsx#L703-L707](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L703-L707)

### Fix

Distinguish "brand has no opportunities at all" (all filters at `all`, zero rows) from "filters matched nothing". In the second case keep the header and filter bar rendered and show a no-matches message with a **Clear filters** button.

### Acceptance criteria

- With active filters and zero matches, the filter bar stays visible and a clear-filters affordance is offered.
- The "No content opportunities yet" + Generate card only appears when the brand truly has no opportunities.
- Unfiltered behavior is unchanged.

lb1192176991-lab:
## What

When the user selects a filter combination that matches zero results (e.g., Status = dismissed on a brand with no dismissed rows), the entire content branch — including the filter bar — was unmounting. This left the user stuck with no way to clear their filters without reloading the page.

## Why

The empty-state gate at  unconditionally rendered the "No content opportunities yet" + Generate card, even when data existed but was hidden by active filters. The filter bar lived only inside the  (results) branch.

## Testing

Now when :
- If any filter is active (status, impact, type, or search), we render the filter bar with a "No opportunities match your filters" message — users can clear individual filters without a page reload.
- If no filters are active, we show the original "No content opportunities yet" + Generate empty state (unchanged behavior).

Closes #660

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
@@ -424,23 +424,172 @@ export default function ContentPage() {
           <Loader2 className="h-8 w-8 animate-spin text-muted-foreground" />
         </div>
       ) : opportunities.length === 0 ? (
-        <Card>
-          <CardContent className="flex flex-col items-center justify-center py-16 gap-4">
-            <Lightbulb className="h-12 w-12 text-muted-foreground/40" />
-            <div className="text-center space-y-1">
-              <p className="text-sm font-medium">{t('noOpportunities')}</p>
-              <p className="text-xs text-muted-foreground">{t('noOpportunitiesHint')}</p>
-            </div>
-            <Button onClick={handleGenerate} disabled={generating} size="sm" className="gap-2">
-              {generating ? (
-                <Loader2 className="h-4 w-4 animate-spin" />
-              ) : (
-                <Lightbulb className="h-4 w-4" />
-              )}
-              {generating ? t('generating') : t('generate')}
-            </Button>
-          </CardContent>
-        </Card>
+        // Show the filter bar even when there are no results, so the user can
+        // clear their filters without reloading the page. The empty-state card
+        // is only appropriate when there are genuinely no opportunities (no
+        // active filters).
+        (() => {
+          const hasActiveFilters =
+            search !== '' ||
+            statusFilter !== 'all' ||
+            impactFilter !== 'all' ||
+            typeFilter !== 'all';
+
+          if (hasActiveFilters) {
+            return (
+              <Card>
+                <CardHeader className="pb-3">
+                  <div className="flex items-center justify-between gap-4 flex-wrap">
+                    <div className="flex items-center gap-2">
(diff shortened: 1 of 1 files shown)


gkhngyk (changes requested):
Thanks for picking this up — the two-state split (filters-matched-nothing vs. genuinely empty) is the right diagnosis. The implementation needs restructuring though:

1. **Don't duplicate the filter bar.** This copy-pastes the entire ~150-line filter Card into the empty branch, so the same UI now exists twice and every future change (e.g. the i18n work happening in #679) has to be made in both places. Instead, lift the filter Card out of the `opportunities.length === 0` conditional so it renders once unconditionally (when the brand has data), and only swap what's below it: table vs. no-matches message vs. generate empty state.

2. **Missing the clear-filters affordance.** The issue's acceptance criteria ask for a "Clear filters" button in the no-matches state — the message alone still leaves an extra step.

3. **CI is red** — the `web — format, lint & typecheck` job fails on this branch; the restructure will likely resolve it, but please make sure it's green before re-requesting review.

The `hasActiveFilters` derivation itself is good and worth keeping.

gkhngyk:
Closing this one — it has gone quiet for nine days, CI is still red, and main has moved 37 commits ahead, so the branch no longer merges. Thanks for the report and the first pass; the diagnosis in your description was right.

We are picking up #660 ourselves. If you would like another one, #656 and #614 are the kind of scope this was, and there are unassigned issues in the tracker.

PR #681, opened 2026-08-07
fix:content search pagination

### LINKED ISSUE #664: Content: search only filters the loaded page and contradicts the pager
### Problem

The search box filters client-side over `opportunities` — which, since server-side pagination, is only the currently loaded page (max 100 rows). The server route supports no search parameter. Two user-facing consequences:

1. On a brand with 340 opportunities, searching scans only the visible 100 — a user searching for a title that lives on page 2 sees "no results" and concludes the record is gone.
2. While a search term is active the pager still renders from the unfiltered `total` — it can read "1–100 of 340" above a table showing 3 rows.

### Evidence

- Client-side filter: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L358-L362](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L358-L362)
- Server route reads only `status, impact, type, limit, offset, sort`: [server/src/routes/content.js#L132](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/content.js#L132)
- Pager bound to `total`: [page.tsx#L708-L715](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L708-L715)

### Fix

Preferred: add a `q` parameter to the list endpoint (`ilike` on title/description), send the debounced search term from the client, and let the pager reflect the searched total. Minimal fallback: scope the UI honestly — hide the pager while a search term is active and change the placeholder to "Search this page…".

### Acceptance criteria

- Searching finds matches beyond the first page (preferred fix), or the UI explicitly says search is page-scoped (fallback).
- The pager numbers never contradict the visible row count while searching.

BharadwajKanneveti:
## Summary

- Added server-side search for content opportunities using the `q` parameter.
- Search now matches opportunity titles and descriptions across all paginated results.
- Debounced the search input to avoid sending a request on every keystroke.
- Reset pagination when search or other filters change.
- Updated the pager to use the total returned for the current search/filter results, preventing contradictory pagination counts.

## Related issue

Closes #664 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/server/src/routes/content.js
+++ b/server/src/routes/content.js
@@ -141,7 +141,7 @@ router.get('/brand/:brandId', async (req, res) => {
     if (type) query = query.eq('type', type);
 
     if (q) {
-      const search = String(q).replace(/([,()])/g, '\\$1');
+      const search = String(q).replace(/[,()]/g, ' ');
       query = query.or(`title.ilike.%${search}%,description.ilike.%${search}%`);
     }
 
--- a/server/src/routes/content.js
+++ b/server/src/routes/content.js
@@ -129,7 +129,7 @@ router.get('/job/:jobId', async (req, res) => {
 router.get('/brand/:brandId', async (req, res) => {
   try {
     const { brandId } = req.params;
-    const { status, impact, type, limit = 50, offset = 0, sort = 'score' } = req.query;
+    const { status, impact, type, q, limit = 50, offset = 0, sort = 'score' } = req.query;
 
     let query = supabaseAdmin
       .from('content_opportunities')
@@ -139,6 +139,11 @@ router.get('/brand/:brandId', async (req, res) => {
     if (status) query = query.eq('status', status);
     if (impact) query = query.eq('impact', impact);
     if (type) query = query.eq('type', type);
+
+    if (q) {
+      const search = String(q).replace(/([,()])/g, '\\$1');
+      query = query.or(`title.ilike.%${search}%,description.ilike.%${search}%`);
+    }
 
     if (sort === 'score') {
       query = query.order('opportunity_score', { ascending: false });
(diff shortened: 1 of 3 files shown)


gkhngyk (changes requested):
This is the preferred fix from #664 — server-side `q` over title/description, the 300ms debounce, and keying the pager on `debouncedSearch` so the page resets and `total` reflects the search. Nice.

One real defect in the escaping: PostgREST's `.or()` logic parser doesn't support backslash escapes, so

```js
const search = String(q).replace(/([,()])/g, '\\$1');
```

doesn't protect anything — a search containing a comma (e.g. `keyword research, planning`) still splits the `or=` expression, PostgREST returns 400, and the user sees "Failed to load". The two working options:

1. Simplest: strip the reserved chars instead of escaping — `String(q).replace(/[,()]/g, ' ')` (they're never meaningful in a fuzzy search anyway), or
2. Proper PostgREST quoting: wrap the pattern in double quotes — `title.ilike."%${escaped}%"` with embedded `"`/`\\` escaped.

Also a heads-up: #682 (same author 🙂) touches the same `loadData`/effect — whichever merges first, the other needs a quick rebase; the `isCancelled` guard there should wrap this PR's search fetches too.

Happy to approve with the escaping sorted.

gkhngyk (approved):
Escaping sorted — stripping the reserved characters is the right call for a fuzzy search, and the rebase onto #682 kept the `isCancelled` guard covering the search fetches too. Debounce, pager reset via the `debouncedSearch` key, and the server-side `q` all line up. Closes #664 properly. Thanks!

PR #682, opened 2026-08-08
fix(web): prevent stale responses from overwriting results

### LINKED ISSUE #663: Content: stale responses can overwrite the list on rapid filter changes
### Problem

`loadData` writes its results unconditionally (`setOpportunities`, `setTotal`), and the effect that calls it has no cleanup. Every filter click starts a new request without cancelling the previous one — so when a user changes filters quickly, a slower earlier response can land **after** the newer one and overwrite the table with rows that don't match the filters on screen. It can also set state after unmount.

### Evidence

- Unconditional writes: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L160-L195](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L160-L195)
- Effect without cleanup: [page.tsx#L197-L199](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L197-L199)
- Accepted pattern to copy elsewhere in the repo: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/%5Bid%5D/page.tsx#L54-L68](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/insights/%5Bid%5D/page.tsx#L54-L68) — `let cancelled = false` + cleanup

### Fix

Guard the state writes with a `cancelled` flag set in the effect's cleanup (the linked insights pattern), so only the latest request may write.

### Acceptance criteria

- Rapidly toggling filters always ends with the table matching the selected filters, regardless of response ordering.
- No state updates after unmount.

BharadwajKanneveti:
## Summary

Prevent stale content opportunity requests from overwriting the table when filters change rapidly.

The content data loader now uses an effect-scoped cancellation flag to ensure that only the latest request can update the opportunities and pagination state. Stale requests are also prevented from changing the loading state after they become outdated.

## Related issue

Closes #663 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
@@ -168,7 +168,7 @@ export default function ContentPage() {
   const pager = usePagination(total, `${statusFilter}|${impactFilter}|${typeFilter}`);
 
   const loadData = useCallback(
-    async (silent = false) => {
+    async (silent = false, isCancelled?: () => boolean) => {
       if (!activeBrandId) {
         setOpportunities([]);
         setTotal(0);
@@ -190,6 +190,7 @@ export default function ContentPage() {
           offset: pager.start,
           sort: 'score',
         });
+        if (isCancelled?.()) return;
         setOpportunities(data.opportunities);
         setTotal(data.total);
         return data.total;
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Clean implementation of #663 — `isCancelled` checked after the await, `setLoading` guarded in `finally`, cleanup in the effect. Approving.

One non-blocking suggestion worth a follow-up commit (here or later): the `catch` still toasts unconditionally, so a request aborted by navigating away can pop "Failed to load content opportunities" on the next page — the same symptom we just filed for Citations in #684. An early `if (isCancelled?.()) return 0;` at the top of the catch closes it.

Heads-up: #681 modifies the same `loadData`/effect — coordinate the merge order between your two branches.

Issue #686, opened 2026-08-08
Prompts: add a date-range selector (7d / 30d / 90d) instead of the hardwired 30-day window


gkhngyk:
### Problem

Every metric on the Prompts page is locked to a 30-day window: the All Prompts table columns (visibility score, mentions, citations, runs) come from `getPromptVisibilitySummaries(brandId, { days: 30 })`, the tooltips hardcode "over the last 30 days", and the Query Fan-out tab passes `days: 30`. Insights and Citations both already offer date presets — Prompts is the only major analytics page without one. The practical gap: after optimizing content, "what changed this week" is invisible inside a 30-day average.

### Evidence

- Fixed window on the summaries call: [web/src/lib/actions/prompt.ts#L553](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/prompt.ts#L553)
- Hardcoded tooltip copy: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/page.tsx#L1871-L1917](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/page.tsx#L1871-L1917)
- Fan-out tab's fixed window: [prompts/_fanout-tab.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/_fanout-tab.tsx) (`getQueryFanout(brandId, { days: 30 })`)
- The backend is already parametric: `getPromptVisibilitySummaries` takes `{ days }`, and the fan-out action accepts `days` capped at `MAX_FANOUT_DAYS = 90`.

### Proposal

Add a preset row — **7d / 30d / 90d**, default **30d** — near the tab bar, applying to all three tabs (All Prompts, Query Fan-out, Insights) so the whole page describes one window and its numbers can't contradict each other.

Design decisions (settled):

- **No 24h preset**: one day = 1–2 runs per prompt; scores would be noise and the table half zeros. **No "All"**: the summaries computation scans every result row, an unbounded window has unbounded cost — and 90d aligns with the fan-out cap.
- **URL-synced**: the page already writes its other filters via `history.replaceState` ([page.tsx#L922](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/page.tsx#L922)) — the range joins those params so views are shareable.
[...]

PR #693, opened 2026-08-10
fix: topic detail trend chart & reports trend now use the AI Visibility Score formula (#685)

### LINKED ISSUE #685: Topic detail: visibility trend chart still uses the legacy raw-average score, contradicting the headline score
### Problem

On the topic detail page the headline **Visibility Score** and the **Visibility Trend** chart come from two different scoring systems:

- The headline uses the current **AI Visibility Score blend** (`computeAiVisibilityScore` — mention/citation/position, same formula as the Insights headline): [web/src/lib/actions/topic.ts#L399-L413](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic.ts#L399-L413)
- The trend chart still calls the **legacy** `getVisibilityTrend`, which averages raw `visibility_score` over ALL answers — including the 0-scored answers where the brand is absent: [web/src/lib/actions/tracking.ts#L3034](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts#L3034) → [tracking.ts#L2088-L2185](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts#L2088-L2185)

The legacy function's own comments admit the problem — *"this JS path is used by the older codepath"* and *"realistic daily averages for a low-visibility brand sit in the 0.3–1.5 band"*. Result: a topic can show a headline score of ~12 while its trend line crawls near 1 — the chart looks broken and understates the topic badly.

### Fix

The new-formula daily series already exists and already supports topic scoping: the `visibility_rate_trend` RPC (migration [00039](https://github.com/ansvisor/ansvisor/blob/main/supabase/migrations/00039_visibility_rate_trend.sql)) takes `p_topic_id` and covers brand + competitors. Point the topic detail's trend (the `getVisibilityTrend(brandId, { topicId })` call in `getTopicDetail`) at that path — the same way the Insights page already consumes it ([tracking.ts#L2246-L2303](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/tracking.ts#L2246-L2303)).

Second legacy caller worth migrating in the same sweep (or a follow-up): the Reports trend section at [web/src/lib/actions/reports.ts#L991](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/reports.ts#L991) — reports pair that trend with new-formula KPIs, so they carry the same mismatch. If both are migrated, `getVisibilityTrend` has no callers left and can be removed. (The MCP/agent surface computes its own aggregates and is out of scope here.)

### Acceptance criteria

- The topic trend chart's latest-day value sits on the same scale as the headline Visibility Score (blend formula), not the raw all-answers average.
- Competitor series on the chart uses the same formula as the brand series.
- Reports trend either migrated alongside or tracked in a follow-up.

Maqbool61:
## Problem

The topic detail page's Visibility Trend chart and the Reports trend section
both called `getVisibilityTrend`, a legacy function that averages raw
`visibility_score` over **all** answers — including 0-scored rows where the
brand was absent. This produced values in the 0.3–1.5 band.

The headline Visibility Score on the same page uses `computeAiVisibilityScore`
(mention/citation/position blend, 0–100 scale). The two numbers were on
completely different scales, making the chart look broken and systematically
understating topic visibility.

Closes: #685

## Fix

**`tracking.ts` — `getTopicDetail`:** call `getVisibilityRateTrend` instead
of `getVisibilityTrend`. The `visibility_rate_trend` RPC (migration 00039)
already accepts `p_topic_id` and covers brand + competitors on the same
formula as the Insights headline. An in-place adapter maps
`VisibilityRateTrendData → VisibilityTrendPoint[]` so `topic_charts.tsx`
needs no changes.

**`reports.ts` — `createReport`:** same swap for the `has('trend')` branch.
Reports pair the trend with new-formula KPIs, so they carried the same
mismatch.

No SQL migration needed — migration 00039 already exists.

## After this PR

`getVisibilityTrend` has zero callers and can be deleted in a follow-up.

## Acceptance criteria
- [x] Topic detail trend chart's latest-day value sits on the same 0–100
  scale as the headline Visibility Score
- [x] Competitor series uses the same formula as the brand series
- [x] Reports trend migrated (or tracked in follow-up)
- [x] No TypeScript errors, lint clean

### DIFF
--- a/web/src/lib/actions/reports.ts
+++ b/web/src/lib/actions/reports.ts
@@ -401,11 +401,7 @@ async function getFanoutSnapshot(
   return [...byQuery.values()]
     .sort((a, b) => b.count - a.count || a.display.localeCompare(b.display))
     .slice(0, REPORT_FANOUT_COUNT)
-    .map((a) => ({
-      query: a.display,
-      engines: [...a.engines].sort(),
-      timesSearched: a.count,
-    }));
+    .map((a) => ({ query: a.display, engines: [...a.engines].sort(), timesSearched: a.count }));
 }
 
 const round1 = (n: number) => Math.round(n * 10) / 10;
@@ -645,11 +641,7 @@ async function getShoppingSnapshot(
   const prev = previousWindow(dateFrom, dateTo);
   const [cur, prior] = await Promise.all([
     getShoppingKpis(brandId, { datePreset: 'all', dateFrom, dateTo }),
-    getShoppingKpis(brandId, {
-      datePreset: 'all',
-      dateFrom: prev.from,
-      dateTo: prev.to,
-    }),
+    getShoppingKpis(brandId, { datePreset: 'all', dateFrom: prev.from, dateTo: prev.to }),
   ]);
 
   const sovPct = round1(cur.shoppingSov * 100);
@@ -729,12 +721,7 @@ async function getReportVisibilityRate(
     visiblePrompts: rate.visiblePrompts,
     promptCount,
     ratePct: promptCount > 0 ? Math.round((rate.visiblePrompts / promptCount) * 1000) / 10 : 0,
-    score: computeAiVisibilityScore({
-      answers,
-      mentionAnswers,
-      citationAnswers,
-      positionFactor,
-    }),
+    score: computeAiVisibilityScore({ answers, mentionAnswers, citationAnswers, positionFactor }),
     mentionAnswers,
     citationAnswers,
     positionFactor,
@@ -999,11 +986,7 @@ export async function createReport(
     has('shareOfVoice') ? getShareOfVoiceData(brandId, range) : null,
     has('competitors') ? getCompetitorComparison(brandId, range) : null,
     has('citations')
-      ? getCitationsOverview(brandId, {
-          datePreset: 'custom',
-          dateFrom,
-          dateTo,
-        })
+      ? getCitationsOverview(brandId, { datePreset: 'custom', dateFrom, dateTo })
       : null,
     has('trend') ? getVisibilityRateTrend(brandId, range) : null,
     has('promptPerformance') ? getPromptPerformance(brandId, dateFrom, dateTo) : null,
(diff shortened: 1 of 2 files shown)


gkhngyk (changes requested):
The core of this is right, and you traced the problem accurately. Both call sites now go through `getVisibilityRateTrend`, the adapter reads the correct `'you'` key, `topicId` and the date range are both supported, and the chart component keeps its existing `VisibilityTrendPoint` shape so nothing downstream had to change. Two things before merge.

**1. The diff carries a lot of unrelated reformatting.** Of ~126 added lines, roughly 18 are the actual fix; the rest are object literals expanded across ~14 hunks in functions this issue never touches — `getPromptResults`, `getVisibilityRateKpi`, `getTrackedPromptsKpi`, `getInsightsFilterOptions`, `getTrackingWindow`, `foldAiVisibility`, `getCompetitorComparison`, `getShareOfVoiceData`, and several spots in `reports.ts`.

This isn't your mistake exactly — Prettier 3.5+ defaults to `objectWrap: "preserve"`, so both the one-line and the expanded form pass `format:check` and CI can't flag it. It's likely your editor expanding them on save. But it makes the real change hard to review, muddies `git blame`, and `tracking.ts` is a file other in-flight work touches (#686 among them), so the extra hunks invite conflicts that buy nothing. Please revert everything except the two call-site changes and their adapters.

**2. `getVisibilityTrend` is now dead code.** After this PR nothing in `web/src` calls it — I checked, the remaining hits are `getVisibilityTrendFor` in the MCP layer, which is a different function and deliberately out of scope. Since the issue's plan was to migrate both callers and drop the legacy function, please delete it here rather than leave an unused legacy path that looks callable.

One thing worth stating in the PR description, since a reviewer will wonder: report snapshots are immutable, so existing reports keep rendering the old-scale values they stored and only new reports use the new formula. That matches how the visibility-score migration handled older snapshots — just make it explicit.

Happy to approve once the diff is down to the real change.

gkhngyk (approved):
Both points addressed cleanly — the diff went from 124/19 to 47/103 and now reads as exactly what it does: two call sites moved to `getVisibilityRateTrend`, two adapters, and the legacy function gone. Thanks for the second pass.

Verified on the branch: `VisibilityTrendPoint` is still exported and used (5 references), and nothing the deleted function depended on became orphaned — `applyModelFilter`, `expandDateToEndOfDay`, `CompetitorMention` and `roundTo1` all still have live callers. Good call adding the snapshot-immutability note in `reports.ts` too.

One leftover, not worth blocking on: the deleted function's JSDoc stayed behind at `tracking.ts:2084-2087` —

```
/**
 * Fetch daily visibility score trend for a brand (and avg competitor score).
 * Queries ALL prompt_results (not deduplicated) to build a time-series.
 */

// ─── Visibility Rate trend (brand + each competitor) ─────────
```

It now reads as documentation for the section below it, which describes the opposite behaviour. If you push once more, drop those four lines; otherwise we'll sweep it and merge as is.

Issue #610, opened 2026-08-03
Content Optimization: list silently caps at 100 opportunities while advertising the full total — wire up the shared TablePager


gkhngyk:
### Problem

The Content Optimization list always fetches with a fixed `limit: 100`, but the UI advertises the full server-side total: the "{total} Available" badge and the Total KPI show numbers the table can never display. Anything past the first 100 opportunities is unreachable from the UI, and the "X shown" sub-label quietly papers over the gap.

### Evidence

- Fixed limit: [`content/page.tsx#L171-L178`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L171-L178) — `getOpportunities(activeBrandId, { ...filters, limit: 100, sort: 'score' })`
- Full total shown anyway: [`#L444`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L444) (`${filtered.length} shown`) and [`#L466-L471`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L466-L471) (`{total} Available` badge)
- The action already supports paging: [`content.ts#L96-L97`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/content.ts#L96-L97) — `limit` and `offset` are passed through as query params
- Shared pager used by the sibling pages: [`table-pager.tsx#L41`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/table-pager.tsx#L41); usage reference [`citations/page.tsx#L303-L310`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L303-L310)

### Suggested fix

Wire up the shared `usePagination`/`TablePager` with server-side paging: pass `offset` from the pager into `getOpportunities`, drive the pager's total from the server `total`, and reset to page 0 when filters change (the citations page shows the resetKey pattern). Bulk-select semantics should be scoped to the current page.

### Acceptance criteria

- All opportunities are reachable through pagination, not just the first 100.
- The badge/KPI totals match what the table can actually show.
- Changing filters resets to the first page.

PR #699, opened 2026-08-10
fix: empty-state Generate button ignores the isCloud restriction

### LINKED ISSUE #661: Content: empty-state Generate button ignores the isCloud restriction
### Problem

The header's manual **Generate** button is deliberately hidden on cloud (`{!isCloud && (<Button onClick={handleGenerate} …>)}`), because opportunities are generated automatically there. But the second Generate button inside the empty-state card renders unconditionally — and the empty state is exactly what a brand-new cloud user sees. They get a manual-generation flow that the product intends to hide.

### Evidence

- Header button correctly gated: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L398](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L398)
- Empty-state button unconditional: [page.tsx#L434-L441](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L434-L441)
- `isCloud` source: [web/src/components/providers/plan-provider.tsx#L35](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/providers/plan-provider.tsx#L35)

### Fix

Wrap the empty-state button in the same `!isCloud` condition. On cloud, show an informational line instead (e.g. "Opportunities are generated automatically as your tracking data comes in.") so the empty state still explains what will happen.

### Acceptance criteria

- Self-host: empty state unchanged, Generate button present.
- Cloud: no Generate button in the empty state; an explanatory sentence appears instead.

kn-dev-code:
## Summary

This PR ensures the empty state "Generate" action respects the `isCloud` plan restriction on the Content Opportunities page. When `isCloud` is true, the manual generate button is replaced with an explanatory message informing the user that opportunities are created automatically from incoming tracking data.

## Related issue

 Closes #661 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist


gkhngyk (changes requested):
The fix itself is correct — the header Generate button is already behind `!isCloud`, the empty-state one wasn't, so it's a real gap. But the PR carries a lot of unrelated change and CI is red on two jobs, so it can't be merged as is.

### Must fix

1. **Revert `supabase/migrations/00048_gsc_suggestion_support.sql`.** That migration has already been applied to the live database. Rewriting it into `CREATE TABLE IF NOT EXISTS` blocks changes what a fresh install produces versus what production actually has, and it's what turns the `schema — consolidated schema.sql is up to date` job red. Shipped migrations are append-only; a correction goes in a new numbered file.
2. **Revert `supabase/config.toml`.** Commenting out `health_timeout` looks like a local workaround for a slow `supabase start`; it isn't part of this fix.
3. **Revert the editor reformatting in `content/page.tsx`.** `catch {}` → `catch { }`, `min-h-[400px]` → `min-h-100`, `w-[130px]` → `w-32.5`, `w-[120px]` → `w-30`, `w-[110px]` → `w-27.5`, the re-indented `impact.*` ternary, and the removed trailing newline at end of file are all unrelated, and together they're what fails `web — format, lint & typecheck`. Please run `yarn format` and `yarn lint` from `web/` and keep the diff to the empty-state block.
4. **Use i18n for the new copy.** `Opportunities are generated automatically as your tracking data comes in.` is a hardcoded English string on a page that goes through `t()`. Add a key (e.g. `content.autoGeneratedHint`) to every file under `web/messages/` and render it with `t('autoGeneratedHint')`.
5. **Link the issue.** The Related issue section is empty — this closes #661, so please add `Closes #661` to the description so the issue closes on merge.

### Note

Once the noise is gone the remaining diff should be roughly the ten lines around the `{!isCloud ? ... : ...}` block, which is exactly what the issue asks for. Ping me when CI is green and I'll take another look.

kn-dev-code:
Thanks for the thorough review and clear feedback! I made some local workaround edits while setting up Supabase locally and accidentally left them in the branch.

I've made the changes per request:

 - Reverted the migration, config, and extra formatting changes.

- Added the en/message for the opportunity message

 - Added i18n support for the new empty state copy.

 - Linked issue #661 in the description.

I'm ready for another look!

gkhngyk (changes requested):
Thanks — the migration and `config.toml` are gone, and the translation key is done properly. Two things left, one of them serious.

### Blocking: the search box stopped working

The latest commits also delete the debounced server-side search that exists on `main`:

- `debouncedSearch` state and the `useEffect` that debounces it
- `filters.q = debouncedSearch.trim()` — so `q` is never sent to the server any more
- `debouncedSearch` from the `usePagination` reset key and from `loadData`'s dependency list

Confirmed against `origin/main`, where all four are present (lines 165, 172, 191–192, 215). Typing in the Content search box now filters nothing — the request goes out without the query, and the page keeps showing whatever the unfiltered page returned.

This looks like the file was restored from a pre-rebase copy rather than the current `main` while reverting the unrelated changes. Rebasing onto `main` and re-applying only the empty-state block should bring it back.

### Prettier is what's failing CI

The `web` job fails on exactly one file, and only on formatting:

```
[warn] src/app/[locale]/(dashboard)/dashboard/content/page.tsx
Code style issues found in the above file.
```

Three leftovers: `catch {}` → `catch { }` in `saveGenerationJob` and `clearGenerationJob`, and the re-indented `impact.high | impact.medium | impact.low` union in the badge. `yarn format` from `web/` fixes all three.

### Also

The Related issue section is still empty — please add `Closes #661` so the issue closes on merge.

The empty-state change itself is right and reads well. Once search is back and CI is green this is good to go.

kn-dev-code:
Alright let's give it another go!

gkhngyk (changes requested):
Search is back and matches `main`, the migration and `config.toml` are reverted, CI is green, and the empty-state fix itself reads correctly. Diffed against `origin/main` rather than the PR view, so this is the real remaining delta — one regression, in the *other* button.

### Blocking: the header Generate button lost three things

Not the empty-state one this issue is about — the one next to Webhook. Against `origin/main`:

| `main` | this PR |
| --- | --- |
| `disabled={generating \|\| loading}` | `disabled={generating}` |
| `RefreshCw` icon once opportunities exist | always `Lightbulb` |
| `t('regenerate')` — "Refresh Opportunities" | always "Generate Opportunities" |

So the button is clickable while the list is still loading, and refreshing an existing set is labelled as generating a new one. `RefreshCw` is now imported and never used, and the `regenerate` translation key has no reader left.

This looks like the same stale-copy problem as last time — the file moved on `main` and the edit was made against an older version. Rebasing onto current `main` and re-applying only the empty-state block should resolve it; nothing about this issue requires touching the header button at all.

### Non-blocking

The cloud hint now also renders in the header, permanently, beside the Webhook button — including when opportunities already exist. In the empty state it explains why there is no button; in the header it is a standing sentence that repeats itself once the empty state is gone. Keeping it to the empty state would read better, but your call.

### Also

The Related issue section still holds the template comment rather than `Closes #661`, so the issue won't close on merge.

The empty-state change is right and the cleanup since the last review is a real improvement — this is the last thing.

kn-dev-code:
Ok @gkhngyk, everything per request should be updated! Please let me know if there are any changes to be made before I resolve conflicts and merge.

gkhngyk (dismissed):
The header button is restored — `disabled={generating || loading}`, the `RefreshCw` icon and the `regenerate` label all match `origin/main` again. Search was already back in the previous round. `Closes #661` is in the description. That is everything I asked for, so this is approved.

I merged it locally against current `main` to check what is left. One conflict, and it is trivial:

```
web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx
```

The `// Server-side paging (#610)` comment lost its indentation in this branch, so it collides with the same lines on `main`. Resolving in favour of `main` is correct — nothing about this fix needs that comment moved. With that one hunk resolved, the merged result passes `typecheck`, `lint`, `build`, and `format:check` after a single `yarn format` (one more comment, `// Restore generation state from localStorage on mount`, also picked up a stray leading space).

So: rebase onto `main`, keep `main`'s version of both comments, run `yarn format` from `web/`, and this is ready.

The empty-state change itself reads well and the `autoGeneratedHint` key is wired properly across the message files. Thanks for sticking with it through three rounds of review — the final diff is exactly the ten lines the issue called for.

kn-dev-code:
Thank you so much @gkhngyk for being patient with me! I really appreciate the guidance you've offered me, and I've learned a lot. If it is ok, I would love to keep assisting with any issues you may have.

gkhngyk (approved):
Rebased cleanly, CI is green, and the net diff is two files — the ten-line empty-state block plus one translation key. Re-approving, since the previous approval was dropped by the rebase.

Verified against current `origin/main` rather than the PR view:

- The header Generate button is byte-identical to `main` — the `loading` guard, the `RefreshCw` icon and the `regenerate` label all survived this round.
- Both paths into `handleGenerate` are now behind `!isCloud`, which is exactly the gap #661 described: the header button always had the guard, the empty-state one never did.
- The `autoGeneratedHint` key is in `messages/en.json`, the only message file in the repo.

The two whitespace hunks (a blank line before the localStorage comment, and the spacing inside `handleBulkDone`) are unrelated to the fix, but they are the shape Prettier settles on and CI's format check agrees, so they are fine to carry.

Thanks for working through the reverts and the rebase — good to merge.

PR #700, opened 2026-08-11
fix: calculate content metrics across all opportunities

### LINKED ISSUE #662: Content: KPI cards mix page-scoped stats with the brand-wide total
### Problem

Since server-side pagination (#610), `opportunities` holds at most one page (`PAGE_SIZE = 100`). Three KPI cards — High Impact, Avg. Score, Sent — are still computed by filtering/averaging that loaded page, while the Total card next to them shows the server-side count of **all** matching rows. On a brand with more than 100 opportunities the row reads e.g. "Total 340 · High Impact 12 · Avg. Score 61" where the last two describe only the first 100 rows — and they silently change when the user pages. Users read these as brand-wide metrics.

### Evidence

- Page-scoped computation: [web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L364-L371](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L364-L371)
- Server total on the adjacent card: [page.tsx#L449](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/content/page.tsx#L449)
- `PAGE_SIZE = 100`: [web/src/components/table-pager.tsx#L12](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/table-pager.tsx#L12)

### Fix (pick one — the first is the good-first-issue version)

1. **Label version:** add "on this page" subtext to the three page-scoped cards so their scope is explicit, and keep Total as-is.
2. **Data version (larger):** extend the list endpoint to return aggregate counts (high-impact count, avg score, sent count) for the whole filtered set and bind the cards to those.

### Acceptance criteria

- KPI numbers can no longer be misread as brand-wide when they aren't: either correctly scoped labels (option 1) or truly brand-wide values (option 2).
- Paging no longer silently changes unlabeled metrics.

BharadwajKanneveti:
## Summary

Fix content opportunity KPIs being calculated only from the current paginated page.

The High Impact, Avg. Score, and Sent metrics are now calculated server-side across the entire filtered dataset, while pagination continues to apply only to the opportunity list. This keeps KPI values consistent when navigating between pages.

## Related issue

Closes #662

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist


gkhngyk (changes requested):
Thanks — moving the KPIs server-side is the right call, and keeping the filter builder in one `applyFilters` helper is clean. One blocking problem before this can go in.

### Blocking: PostgREST aggregate functions are not enabled

`avgScoreQuery` uses `select('opportunity_score.avg()')`. Aggregate functions in `select` are a PostgREST feature that is off unless `db-aggregates-enabled` is turned on, and it is off on our deployment. Verified against the live REST endpoint:

```
GET /rest/v1/content_opportunities?select=opportunity_score.avg()
400 {"code":"PGRST123","message":"Use of aggregate functions is not allowed"}
```

Because `avgError` is rethrown, this doesn't just zero the KPI — the whole `GET /api/content/brand/:brandId` returns 500, so the Content page stops loading entirely. It can't be caught in review by CI since nothing exercises the route against a real PostgREST.

Suggested fix: compute the three aggregates in a single Postgres function and call it with `supabaseAdmin.rpc(...)`. That also collapses the current four round-trips into two — the list plus one aggregate call — which matters here because the aggregates re-scan the filtered set on every page change. Something like:

```sql
create function content_opportunity_aggregates(
  p_brand_id uuid, p_status text, p_impact text, p_type text, p_q text
) returns table (avg_score numeric, high_impact_count bigint, sent_count bigint)
```

with the same filter semantics as `applyFilters`, added as a numbered migration under `supabase/migrations/`.

If you'd rather keep it in JS, the two `head: true` counts are fine as they are — only the average needs another route.

### Non-blocking

- `setAggregates(data.aggregates)` trusts the field to exist. A server that hasn't been redeployed yet returns no `aggregates`, and the KPI cards then read `.avgScore` off `undefined` and crash the page. Worth defaulting: `setAggregates(data.aggregates ?? { avgScore: 0, highImpactCount: 0, sentCount: 0 })`.
- `highImpactQuery` is built even when `impact` is set to something other than `high`, and then never awaited. It's harmless (the builder only runs when awaited), but building it inside the `if` would make the intent obvious.
- The diff reformats the untouched `.order(...)` calls onto three lines each and drops the blank line before the next route handler. Reverting that noise would make the real change much easier to read.

Semantics all look right otherwise: `status` + `.in('status', [...])` correctly yields 0 when the user filters to a non-sent status, and `impact=high` doesn't double-apply the filter.

gkhngyk (changes requested):
This is much better — the RPC is clean, `SECURITY INVOKER` with a grant to `authenticated` is the right call (RLS on `content_opportunities` still applies to a direct caller), four round trips are down to two, and the `data.aggregates ?? …` fallback is in. I checked the aggregate logic against production data and it produces exactly what the issue describes: 766 opportunities for one brand, 372 high-impact, 100 sent — where the old page-scoped calculation could only ever count the 50 rows on screen.

One real fix left.

### The search string diverges between the list and the KPIs

The data query sanitises `q`, the RPC does not:

```js
// list
const search = String(q).replace(/[,()]/g, ' ');
filteredQuery.or(`title.ilike.%${search}%,description.ilike.%${search}%`);

// aggregates
p_q: q ? String(q) : null,
```

So for any query containing `,`, `(` or `)` the two halves of the response describe different result sets — the list searches `foo  bar` while the KPI cards search `foo, bar`. That is the exact class of mismatch this PR exists to remove, just moved from pagination to search.

The stripping is a PostgREST filter-syntax workaround and has no reason to apply to a bound function parameter, but the two must agree. Simplest is to compute it once and pass the same value to both:

```js
const search = q ? String(q).replace(/[,()]/g, ' ') : null;
```

then use `search` inside `applyFilters` and as `p_q`. That keeps today's list behaviour unchanged and makes the aggregates match it exactly.

### Two notes, neither blocking

- **Migration number collision.** #703 also adds a `00050_…`. Whichever merges second needs renumbering to `00051` and a `bash supabase/build-schema.sh` re-run. Nothing wrong on your side — just so it isn't a surprise.
- The migration file has no trailing newline. `build-schema.sh` compensates so `schema.sql` is still valid, but it makes the generated file's spacing inconsistent with every other block.

Ping me once the search value is shared and I'll take another look.

gkhngyk (approved):
Both points addressed exactly — `search` is computed once and feeds `applyFilters` and `p_q`, so the list and the KPI cards can no longer describe different result sets, and the migration ends with a newline. CI is green across all jobs.

Approving. Thanks for the quick turnaround on both rounds — the RPC version is a better piece of code than what the issue asked for.

One thing to expect on merge, not a change request: #703 also adds a `00050_…` migration. Whichever lands second gets renumbered to `00051` with a `bash supabase/build-schema.sh` re-run. I'll handle it on my side if this merges first.

PR #714, opened 2026-08-13
open the All Prompts list sorted by visibility score, highest first

### LINKED ISSUE #712: Prompts: open the All Prompts list sorted by visibility score, highest first
### Problem

The All Prompts table opens sorted by creation date, newest first. That is a convenient default, not a useful one: after the first week it tells the user nothing, and it answers a question nobody is asking on a visibility product's prompt list.

The table already sorts by visibility — the column header is clickable and the choice persists to the URL — so today every user who wants the obvious view has to click to get it, on every fresh visit that doesn't carry the query string.

Default to visibility, highest score first, so the list opens as a leaderboard: the prompts the brand is strongest on at the top, weakest further down.

### Where it lives

- Default ordering: [prompts/page.tsx](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/page.tsx) — the `if (!activeSort)` branch in `filtered`, currently `createdAt` descending
- Sort state and URL sync: same file, `activeSort` / `dir` read from `searchParams`
- Comparator: [prompt-sort.ts](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/prompts/prompt-sort.ts)

### The part that is easy to get wrong

**The header indicator has to agree with the default.** `activeSort` is `null` when the URL carries no `sort` parameter, and the sort arrow only renders on the active column. Changing just the fallback ordering would leave the table sorted by visibility while every header still reads as unsorted — the user clicks Visibility expecting something to happen and gets the same order back. So the default has to make `visibility` the active sort, not merely the fallback comparison.

Re-clicking must still toggle to ascending, and the URL should keep behaving as it does now.

### A deliberate consequence, not a bug

`compareNullsLast` sends null values to the bottom in *both* directions, on purpose — ascending shouldn't fill the top of the table with blank rows. A prompt that has never been tracked has no visibility score, so under this default newly added prompts sort to the bottom until their first run.

That is the correct behaviour and should stay, but it is worth knowing: someone who adds twenty prompts and reloads will find them at the end of the list rather than the top, which the previous newest-first default gave them for free. If that turns out to bother people in practice, the fix is a separate conversation about surfacing untracked prompts, not a change to the comparator.

### Acceptance criteria

- Opening the All Prompts tab with no query string sorts by visibility score, highest first.
- The Visibility header shows itself as the active sort in that state.
- Clicking Visibility toggles to ascending; clicking another column behaves exactly as it does today.
- A shared or reloaded URL that already carries `sort` and `dir` still wins over the default.
- Prompts with no visibility score remain at the bottom in both directions.

Pavel-glitch-ui:
- Default `activeSort` state to 'visibility' (descending) when no `sort` searchParam is present in URL
- Ensure Visibility column header displays as active with descending chevron indicator on initial load
- Remove fallback sorting by `createdAt` in favor of unified `compareNullsLast` comparator
- Preserve bottom placement for un-tracked prompts (null visibility scores)
- Maintain explicit URL searchParam precedence and toggle-to-asc behavior on header re-click

Closes #712

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/prompts/page.tsx
@@ -1761,7 +1761,7 @@ function AllPromptsTab({
   );
 
   const rawSort = searchParams.get('sort');
-  const activeSort: AllPromptsSortKey | null = isAllPromptsSortKey(rawSort) ? rawSort : null;
+  const activeSort: AllPromptsSortKey = isAllPromptsSortKey(rawSort) ? rawSort : 'visibility';
   const dir: SortDir = searchParams.get('dir') === 'asc' ? 'asc' : 'desc';
 
   // Click a header to sort: a new column starts at desc ("show me the extremes");
@@ -1795,11 +1795,6 @@ function AllPromptsTab({
       );
     }
 
-    if (!activeSort) {
-      // Default ordering: newest first.
-      return list.sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime());
-    }
-
     const valueOf = (p: Prompt): number | null => {
       const vis = visibility[p.id];
       const vol = volumeByPromptId.get(p.id);
@@ -1826,7 +1821,7 @@ function AllPromptsTab({
   // changes the visible row set (search text, sort column, sort direction, work
   // status filter) so usePagination jumps back to page 0 automatically when
   // any filter or sort changes — satisfying the issue's acceptance criteria.
-  const promptsResetKey = `${search}::${workFilter}::${activeSort ?? ''}::${dir}`;
+  const promptsResetKey = `${search}::${workFilter}::${activeSort}::${dir}`;
   const promptsPager = usePagination(filtered.length, promptsResetKey);
 
   if (loading) {


gkhngyk (approved):
Correct and minimal — four lines of real change, and it lands on the part of the issue that was easy to get wrong.

That part being the header indicator. Making only the fallback comparison sort by visibility would have left the table visibly ordered while every header still read as unsorted, so the first click on Visibility would appear to do nothing. Defaulting `activeSort` itself rather than the comparator is exactly right: `SortableHead` computes `active` from it, so the chevron and `aria-sort="descending"` come along for free, and `handleSort` still toggles to ascending on re-click because it compares against the same value.

Things I checked beyond the diff:

- **No reshuffle on load.** I wondered whether the table could render with `visibility` still empty — every score would read as null, sort would be a no-op, and the rows would jump once the data arrived. It can't: `setAllPrompts` and `setVisibility` are assigned together and `loading` only clears afterwards, so the table never renders against a half-populated state.
- **Ties behave sensibly.** `compareNullsLast` returns 0 for equal values and `Array.prototype.sort` is stable, so prompts on the same score keep the order they arrived in — which is `created_at` descending from the server. Newest-first survives as the implicit tiebreak rather than being lost outright.
- **URL precedence and null placement** are untouched, as intended: an explicit `?sort=`/`?dir=` still wins, and untracked prompts stay at the bottom in both directions.
- `yarn typecheck`, `format:check`, `lint` and `test` all clean locally; the six lint warnings are pre-existing elsewhere in the tree.

Two non-blocking notes, neither worth another push:

`SortableHead` still types `activeSort` as `AllPromptsSortKey | null`. Harmless, but the null case is now unreachable and tightening it would document that.

Sorting by date is no longer reachable at all, since there is no `createdAt` sort key — so a user who has just added twenty prompts can't bring them to the top. That is the consequence the issue signed off on, and the right answer if it turns out to bother people is surfacing untracked prompts rather than restoring the old default. Flagging it only so it isn't mistaken for a regression later.

Approving. Thanks for keeping the diff this tight.

PR #742, opened 2026-08-19
fix(citations): replace source filters with client-side scope

### LINKED ISSUE #655: Citations: replace the three interlinked category toggle buttons with a single source-scope selector
### Problem

The Citations filter bar has three separate toggle buttons for source category — **Exclude own domain**, **Competitors only**, and **Own domain only** — that actually represent one mutually-exclusive choice. They keep each other consistent through hand-wired `onChange` logic: turning on *Competitors only* silently also enables *Exclude own domain*, each button clears the others, and so on. The result reads as overlapping/redundant controls ("Exclude own domain" and "Competitors only" look like the same thing), and the toggle-button widget suggests the options are freely combinable when they aren't.

The wording is also narrower than what the page shows: cited sources aren't only domains.

### Evidence

- Toggle buttons + cross-toggle wiring: [web/src/components/citations/filter-bar.tsx#L362-L408](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/citations/filter-bar.tsx#L362-L408) — note `excludeOwnDomain: !filters.competitorOnly ? true : …` inside the *Competitors only* handler
- Three booleans in filter state: [web/src/components/citations/filter-bar.tsx#L47-L49](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/citations/filter-bar.tsx#L47-L49)
- Filter application (the three flags are mutually exclusive in effect): [web/src/lib/actions/citations.ts#L295-L297](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts#L295-L297)

### Proposal

Replace the three buttons with a single **Sources** selector (segmented control or `Select`, matching the other filter-bar controls) with four options:

| Option | Behavior (today's equivalent) |
|---|---|
| **All sources** | no category filter (default) |
| **Third-party only** | hide the brand's own sources — competitors *and* neutral sources stay (today's *Exclude own domain*) |
| **Competitor sources** | only competitor sources (today's *Competitors only*) |
| **Own sources** | only the brand's own sources (today's *Own domain only*) |

Model it as one field, e.g. `sourceScope: 'all' | 'third_party' | 'competitors' | 'own'`, replacing the `excludeOwnDomain` / `competitorOnly` / `ownOnly` booleans in the filter state, the server action's filter params, and everywhere the page threads them through. The hand-wired mutual-exclusion logic disappears entirely — inconsistent combinations become unrepresentable.

### Acceptance criteria

- One selector labeled **Sources** with the four options above; the three toggle buttons are gone.
- Each option filters exactly as its old counterpart did (*Third-party only* must keep neutral sources visible — it is not the same as *Competitor sources*).
- Filter state carries a single scope value; the `excludeOwnDomain`/`competitorOnly`/`ownOnly` booleans and their cross-toggle wiring are removed end to end (filter bar, page, server action).
- Reset/clear-filters returns the selector to **All sources**.

### LINKED ISSUE #657: Citations: source-scope changes re-scan the whole window server-side even though rows are already classified
### Problem

Loading the Citations page is expensive by design: `getCitationsOverview` pages through every `prompt_results` row in the selected window (1000 rows per sequential PostgREST request, up to 50k rows) and classifies every citation in the server action. That cost is acceptable for a genuine window change — but today it is also paid when it buys nothing:

- Toggling a source-category filter (*Exclude own domain* / *Competitors only* / *Own domain only* — soon the single **Sources** selector from #655) triggers a full re-scan and re-download of the exact same rows. The category flags don't affect the scan at all; they only skip rows during aggregation.
- Opening the **Competitor Gaps** tab runs a second full scan of the same window.

On large brands with wide date ranges this turns a cheap presentational choice into a multi-second wait.

### Evidence

- Sequential scan loop: [web/src/lib/actions/citations.ts#L149-L214](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts#L149-L214)
- Category flags applied only at aggregation time: [citations.ts#L295-L297](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts#L295-L297) — the fetched data is identical with or without them
- Client refetches on every category-flag change because the flags are part of `apiFilters`: [web/src/app/%5Blocale%5D/(dashboard)/dashboard/citations/page.tsx#L858-L871](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L858-L871)
- Competitor Gaps second scan: [citations.ts#L623](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/citations.ts#L623)

### Proposal

Make source-scope a client-side concern. The overview payload already carries what's needed: every domain/URL row has a `category` (`you` / `competitor` / neutral), and `sourceTypeBreakdown` exists. Concretely:

- Remove the category flags from the server action's filters. The action returns the full (scope-unfiltered) aggregation for the window.
- Apply the selected scope (#655's `sourceScope`) client-side: filter domain/URL rows by `category` and derive the KPI totals for the active scope from per-category counts included in the payload.
- Result: switching **Sources** is instant — zero network, zero re-scan. Only date/platform/topic/prompt/region changes (which genuinely change the scanned window) refetch.

This pairs naturally with #655 — implementing both together avoids wiring the new selector into the server round-trip only to remove it again.

### Future work (out of scope here)

- Fetch scan pages concurrently instead of sequentially, or push the whole aggregation into a Postgres function so only aggregates cross the wire.
- Reuse the overview scan for Competitor Gaps instead of scanning the window twice.

### Acceptance criteria

- Changing the source scope updates KPIs, tables, and breakdowns without any server request.
- Numbers for each scope are identical to today's server-filtered results.
- Window-changing filters (date, platform, topic, prompt, region) still refetch as before.

BharadwajKanneveti:
## Summary

- Replaced the three mutually-exclusive citation source toggles with a single `Sources` selector:
  - All
  - Brand
  - Competitors
  - Third-party
- Added `sourceScope` to the UI filter state and removed the old `excludeOwnDomain`, `competitorOnly`, and `ownOnly` fields and wiring.
- Moved source-scope filtering to the client using the existing citation `category` values, avoiding a backend re-scan when the source scope changes.
- Applied source-scope filtering to both domain and URL tables.
- Updated pagination and CSV exports to use the filtered domain and URL rows.
- Kept KPI totals based on the overall citation window rather than the selected source scope.
- Preserved the distinction between `Third-party` and `Competitors`: third-party includes the six non-brand, non-competitor categories.

## Related issue

Closes #655
Closes #657

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist


BharadwajKanneveti:
@gkhngyk I have made a draft pr as you asked please check and let me know if you find any issues.

gkhngyk (approved):
Thanks — this matches the design decision exactly: a strict partition, with third-party meaning neither brand nor competitor rather than everything-except-mine. Moving the scope client-side removes the re-scan #657 was about, and pagination and the CSV export both follow the filtered rows rather than the raw ones.

One interaction to flag, and it is mine, not yours: #744 landed yesterday and caps the URL list at the top 2,000 by citation count. Your filter applies within that set, so scoping to Competitors can under-report URLs whose competitors sit below the cap. Domains are unaffected — those are returned uncapped. Fixing it means teaching the URL function about scope so the cap applies inside it, which is a separate change; I'll open an issue and take it.

Nothing to change here.

PR #758, opened 2026-08-21
fix(csv): quote fields containing lone carriage returns

### LINKED ISSUE #757: bug: CSV export does not quote a field containing a lone carriage return
### Description

`escape()` in [`web/src/lib/csv.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/csv.ts) decides whether a field needs quoting with `/[",\n]/`. That covers the quote, the comma and the line feed, but not a **lone carriage return**.

RFC 4180 requires a field containing CR *or* LF to be quoted. A parser reading an unquoted CR treats it as the end of a record, so one such value silently splits a row in two and shifts every column after it.

CRLF happens to be safe — the `\n` in it matches, so the field gets quoted. Only a bare `\r` slips through, which makes this narrow, but it is reachable: exported values include prompt text, brand names and citation titles, and scraped or pasted content can carry a bare CR.

The same function is behind every CSV in the product — Insights, Citations, Topics, Prompts, Traffic and Shopping all call `toCsv`.

### Steps to reproduce

1. Call `toCsv([{ text: 'a\rb' }], ['text'])`.
2. Look at the field it produces.

Or, in the product: get a value containing a bare CR into an exported column and open the file in a spreadsheet.

### Expected behavior

The field is quoted, the same way a value containing `\n` already is:

```
text
"a\rb"
```

### Actual behavior

The field is emitted raw, with the CR still in it:

```
text
a\rb
```

A parser reads that as two records. Every column after the break is shifted.

### Deployment mode

Cloud (ansvisor.com)

### Environment

Not environment-specific — `toCsv` is a pure function.

The fix is in the character class. Make sure `\r` alone triggers quoting and that CRLF keeps behaving as it does now.

[`web/src/lib/csv.test.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/csv.test.ts) already covers the comma, quote and newline cases in the style to follow. Add cases for a bare `\r` and for `\r\n`, and check that the new test fails before the fix and passes after — a test that passes either way proves nothing.

**Out of scope:** values beginning with `=`, `+`, `-` or `@` are also passed through unchanged and a spreadsheet will evaluate them as formulas. That is a real concern but a separate one, and mitigating it changes the exported value, so it needs its own decision. Please don't fold it in here.

dawNotPoi:
## Summary

- Quote CSV fields containing a lone carriage return so parsers do not split one row into multiple records.
- Preserve the existing quoting behavior for CRLF values.
- Add regression coverage for both lone CR and CRLF inputs.

## Related issue

Closes #757

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Testing

- `yarn test` — 16 files, 90 tests passed
- `yarn typecheck`
- `yarn lint` — 0 errors (8 pre-existing warnings)
- `yarn format:check` — passed in a clean LF checkout matching CI
- `yarn build`
- `git diff --check`

## Checklist

### DIFF
--- a/web/src/lib/csv.test.ts
+++ b/web/src/lib/csv.test.ts
@@ -21,6 +21,16 @@ test('newline in a value triggers quoting', () => {
   expect(result).toBe('text\n\"line1\nline2\"');
 });
 
+test('lone carriage return in a value triggers quoting', () => {
+  const result = toCsv([{ text: 'line1\rline2' }], ['text']);
+  expect(result).toBe('text\n\"line1\rline2\"');
+});
+
+test('CRLF in a value remains quoted', () => {
+  const result = toCsv([{ text: 'line1\r\nline2' }], ['text']);
+  expect(result).toBe('text\n\"line1\r\nline2\"');
+});
+
 test('null and undefined become empty strings', () => {
   const result = toCsv([{ a: null, b: undefined, c: 'value' }], ['a', 'b', 'c']);
   expect(result).toBe('a,b,c\n,,value');
--- a/web/src/lib/csv.ts
+++ b/web/src/lib/csv.ts
@@ -2,7 +2,7 @@ function escape(value: unknown): string {
   if (value === null || value === undefined) return '';
 
   const s = String(value);
-  if (/[",\n]/.test(s)) return `"${s.replace(/"/g, '""')}"`;
+  if (/[",\r\n]/.test(s)) return `"${s.replace(/"/g, '""')}"`;
   return s;
 }
 


gkhngyk (approved):
Thanks — this is exactly the fix.

Verified the character class change does what the issue asked: `line1\rline2` now quotes where it previously went out raw, and CRLF, LF, comma, quote and plain values are all unchanged. The lone-CR test fails on the old regex and passes on the new one, so the coverage is real.

Scope is right too — the formula-injection case stays untouched.

CI green across the board.

PR #763, opened 2026-08-21
Fix(slugify): transliterate non ascii

### LINKED ISSUE #760: bug: slugify deletes non-ASCII letters instead of transliterating them
### Description

`slugify` throws away every character outside `a-z0-9-`:

```ts
.replace(/[^a-z0-9-]/g, '')
```

Accented and non-ASCII letters are not transliterated, they are **deleted**. For a product whose brands are often Turkish, that hollows the name out rather than shortening it:

| Name | Slug produced |
|---|---|
| `Ürün` | `rn` |
| `Şeker` | `eker` |
| `Çiğdem` | `idem` |
| `Öz Güven` | `z-gven` |
| `Ağaç` | `aa` |

The same applies to German, French, Nordic and Slavic names — `Müller` becomes `mller`.

Nothing routes on the slug, so this breaks no navigation. Where it shows up is **filenames**: every CSV and PDF export is named `ansvisor_${slug}_…`, so a brand exports as `ansvisor_rn_prompts_2026-08-21.csv`. It also collides more eagerly than it should — `Ağaç` and `Aa` produce the same slug.

There are **three copies** of this logic, and they have already drifted apart in their fallbacks:

- [`web/src/lib/actions/brand.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/brand.ts) — falls back to `brand-${Date.now()}`
- [`web/src/lib/actions/organization.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/organization.ts) — falls back to `org-${Date.now()}`
- [`onboarding/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(onboarding)/dashboard/onboarding/page.tsx) — the same chain written inline, no shared function at all

### Steps to reproduce

1. Create a brand named `Ürün`.
2. Look at its `slug`, or export any CSV from one of its pages.

### Expected behavior

The slug carries the name across: `urun`. Exports read `ansvisor_urun_prompts_….csv`.

### Actual behavior

The slug is `rn`, and exports read `ansvisor_rn_prompts_….csv`.

### Deployment mode

Cloud (ansvisor.com)

### Environment

Not environment-specific.

**Transliterate before stripping.** The tidy way is Unicode normalization — `normalize('NFD')` splits a letter from its accent so the accent can be removed on its own, which handles `é → e`, `ü → u` and most of Latin-1 without a lookup table.

Two Turkish letters do **not** decompose and need an explicit mapping: `ı` and `ş`, plus `ğ`. Handle those before or alongside the normalization pass, and take care with `I`/`ı` — lowercasing Turkish is famously not the same as lowercasing English, so test `Işık` explicitly.

**Please also pull the three copies into one shared helper** while you are in there. Three implementations of the same rule is how they drifted in the first place; a single one with tests is the real fix. There is no test file for it today — a new one alongside [`web/src/lib/csv.test.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/csv.test.ts) is the pattern to follow.

**Do not backfill existing rows.** Slugs are stored, and rewriting them is a separate decision with its own consequences. This issue is about what new names produce.

shahzainshafique:
## Summary

slugify was stripping every character outside [a-z0-9-], which deleted accented and non-ASCII letters instead of transliterating them. Ürün came out as rn, so CSV/PDF exports were named ansvisor_rn_…, and it collided more than it should (Ağaç and Aa both produced aa). Given how many brands are Turkish, this hollowed names out fairly often.

This PR transliterates before stripping. An explicit map handles the letters Unicode NFD doesn't decompose on its own (Turkish ı/ş/ğ, German ß, Nordic ø, Polish ł, etc.) and pins the Turkish I/ı lowercasing case to i. An NFD pass then covers the rest of Latin-1 (é→e, ü→u, ç→c). Ürün now slugifies to urun.

It also pulls the three drifted copies of this logic in brand.ts, organization.ts, and the inline chain in onboarding/page.tsx and turns them into one shared helper at web/src/lib/slug.ts, with each call site keeping its own fallback prefix (brand-, org-). New unit tests sit next to csv.test.ts and include Işık for the Turkish lowercase gotcha. Existing slugs are intentionally not backfilled.

## Related issue

Closes #760

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx
+++ b/web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx
@@ -9,6 +9,7 @@ import { getPromptCapacity, savePromptSet } from '@/lib/actions/prompt';
 import { triggerTrackingCheck } from '@/lib/actions/tracking';
 import { addCompetitor, getCompetitors } from '@/lib/actions/competitor';
 import { getFaviconUrl } from '@/lib/favicon';
+import { slugify } from '@/lib/slug';
 import { useBrandStore } from '@/stores/use-brand-store';
 import { REGIONS, US_STATES, LANGUAGES } from '@/config/prompt-options';
 import { ALL_MODELS, ALL_SCRAPERS } from '@/config/prompt-options';
@@ -511,14 +512,7 @@ export default function OnboardingPage() {
       let orgId = profile?.organization_id;
 
       if (!orgId) {
-        const slug =
-          brandName
-            .toLowerCase()
-            .trim()
-            .replace(/\s+/g, '-')
-            .replace(/[^a-z0-9-]/g, '')
-            .replace(/-+/g, '-')
-            .replace(/^-|-$/g, '') || `org-${Date.now()}`;
+        const slug = slugify(brandName) || `org-${Date.now()}`;
 
         const { data: existing } = await supabase
           .from('organizations')
--- a/web/src/lib/actions/brand.ts
+++ b/web/src/lib/actions/brand.ts
@@ -3,17 +3,8 @@
 import { createClient } from '@/lib/supabase/server';
 import { revalidatePath } from 'next/cache';
 import { enforceLimit } from '@/lib/guards/plan-guard';
+import { slugify } from '@/lib/slug';
 import type { Brand, BrandDomain } from '@/types';
-
-function slugify(text: string): string {
-  return text
-    .toLowerCase()
-    .trim()
-    .replace(/\s+/g, '-')
-    .replace(/[^a-z0-9-]/g, '')
-    .replace(/-+/g, '-')
-    .replace(/^-|-$/g, '');
-}
 
 function mapDomainRow(d: Record<string, unknown>): BrandDomain {
   return {
--- a/web/src/lib/actions/organization.ts
+++ b/web/src/lib/actions/organization.ts
@@ -2,16 +2,7 @@
 
 import { createClient } from '@/lib/supabase/server';
 import { revalidatePath } from 'next/cache';
-
-function slugify(text: string): string {
-  return text
-    .toLowerCase()
-    .trim()
-    .replace(/\s+/g, '-')
-    .replace(/[^a-z0-9-]/g, '')
-    .replace(/-+/g, '-')
-    .replace(/^-|-$/g, '');
-}
+import { slugify } from '@/lib/slug';
 
 export async function createOrganization(name: string) {
   const supabase = await createClient();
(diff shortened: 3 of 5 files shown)


gkhngyk (changes requested):
Thanks for taking this on — the transliteration itself is right, and the consolidation is what the issue asked for. I pulled the branch and ran it: the 8 new tests pass, `tsc --noEmit` is clean, lint has 0 errors, all three copies are gone with no leftovers, and each call site keeps its own fallback prefix. Every case from the issue's table comes out right, including `Işık` and `İstanbul`.

Three things before merge.

**1. `format:check` fails — blocking.**

CI is red on it, and it is one trailing space at the end of `src/lib/slug.ts:31`:

```
 * brand's name.·
```

Worth correcting the note in the PR description: this is not pre-existing. I ran `yarn format:check` from `web/` across the whole tree and `src/lib/slug.ts` is the only file that fails — the four files you edited are all Prettier-clean, and `server` and `schema` passed. `yarn format` from `web/` fixes it.

**2. The character list is written twice.**

```ts
.replace(/[ıİIşŞğĞßøØłŁđĐðÐþÞæÆœŒ]/g, (ch) => TRANSLITERATIONS[ch] ?? ch)
```

The same 22 characters live in the regex and in the map's keys. They agree today, but this issue existed because one rule was written in three places and they drifted — reintroducing that on a smaller scale is worth avoiding. Deriving the class from `Object.keys(TRANSLITERATIONS)`, or just matching `[^\u0000-\u007F]` and letting `?? ch` pass unmapped characters through to the NFD pass, removes the second copy.

**3. Nit:** `test('plain ASCII names lowercase and hyphenate')` asserts `slugify('Öz Güven')`, which is not ASCII. Either rename it or move that case in with the accented ones.

Non-blocking and out of scope, just noting it: a name written entirely in a non-Latin script still slugs to empty and falls back to the timestamp. That is unchanged from before, so nothing regressed.

shahzainshafique:
Thanks for the thorough review! Fixed all three:

- Trailing space is gone: format:check passes now. And you're right, my note was wrong; I've corrected it, that file was the only real failure.
- Good call on the duplication: that's exactly the trap this issue was about. The character set now lives only in the map and the regex is derived from its keys via Object.keys, so the two can't drift.
- Renamed the ASCII test to use actual ASCII and moved Öz Güven in with the accented cases.

Squashed it down to a single commit. The non-Latin-script-to-empty fallback is unchanged as you noted.

gkhngyk (approved):
All three addressed, CI green. Deriving the pattern from the map's keys is the right call. Thanks!

PR #772, opened 2026-08-22
fix(#767): consolidate relative-time formatting into a single shared helper

### LINKED ISSUE #767: bug: four copies of the relative-time helper disagree about the same timestamp
### Description

"Last run" timestamps are rendered by four separate copies of the same helper, under three different names, and they have drifted apart:

| Where | Name | Under a minute | Rounding | Past 30 days |
|---|---|---|---|---|
| [topics/page.tsx#L57](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/topics/page.tsx#L57) | `formatRelative(iso, t)` | `t('relative.justNow')` | `Math.floor` | locale date |
| [prompts/page.tsx#L274](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/prompts/page.tsx#L274) | `formatRelative(iso?)` | `Just now` | `Math.round` | locale date |
| [insights/page.tsx#L1731](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/insights/page.tsx#L1731) | `formatTimeAgo(date: Date)` | `just now` | `Math.floor` | `Xd ago` |
| [traffic/page.tsx#L203](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/traffic/page.tsx#L203) | `timeAgo(dateStr)` | `just now` | `Math.floor` | `Xd ago` |

Three of those differences are visible to the user:

- **Rounding.** A run 90 seconds ago reads `2m ago` on Prompts and `1m ago` on Topics. At 45 minutes Prompts says `1h ago` while the other three still say `45m ago`.
- **Casing.** `Just now` on Prompts, `just now` everywhere else.
- **The 30-day cliff.** Topics and Prompts hand off to a locale date; Insights and Traffic keep counting days indefinitely, so a year-old timestamp reads `412d ago`.

Only the Topics copy goes through i18n.

### Steps to reproduce

1. Note a brand's last-run time on the Topics page.
2. Open the Prompts page and find the same brand's last run.
3. Compare them at a point where the rounding differs — 90 seconds, or 45 minutes.

### Expected behavior

One timestamp renders one way, wherever it appears.

### Actual behavior

The same instant produces different strings depending on which page you are on.

### Deployment mode

Cloud (ansvisor.com)

### Environment

Not environment-specific — these are pure functions.

**The Topics copy is the one to keep.** It uses `Math.floor`, hands off to a locale date past 30 days, and reads its strings from `topics.relative.*` in `web/messages/en.json`.

- New home: `web/src/lib/format-relative.ts`, with `web/src/lib/format-relative.test.ts` beside it. [`web/src/lib/csv.test.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/csv.test.ts) is the pattern to follow for both.
- The `relative.*` keys currently live under the `topics` namespace. Moving them to `common` — which already exists — lets all four call sites share them.
- The Insights copy takes a `Date`; the other three take an ISO string. Pick one signature and adjust that call site rather than supporting both.
- Prompts' copy returns `—` for a missing value; keep that, the table depends on it.
- Adopting the Topics behaviour changes what Insights and Traffic show past 30 days. That is the intent, but call it out in the PR so it doesn't read as an accident.

Tests worth having: the sub-minute boundary, `floor` vs `round` at 90 seconds, the hour and day boundaries, the 30-day cliff, and a null/undefined input.

Maqbool61:
## Summary

Closes #767 

Four pages (`topics`, `prompts`, `insights`, `traffic`) each had their own
local implementation of the same relative-time helper. They had drifted into
three user-facing bugs:

| Bug | Where |
|---|---|
| `Math.round` instead of `Math.floor` → 90s showed as "2m ago", 45min as "1h ago" | `prompts/page.tsx` |
| Hardcoded `'Just now'` (capital J) instead of `'just now'` | `prompts/page.tsx` |
| No 30-day cliff → timestamps showed as "412d ago" indefinitely | `insights/page.tsx`, `traffic/page.tsx` |
| Hardcoded English strings, no i18n | `insights/page.tsx`, `traffic/page.tsx` |

## Changes

- **`web/src/lib/format-relative.ts`** — single canonical implementation.
  `Math.floor` throughout, 30-day cliff falls back to `toLocaleDateString()`,
  i18n via `common.relative.*` keys.
- **`web/src/lib/format-relative.test.ts`** — 14-case test suite covering
  null/undefined, sub-minute, the 90s floor-vs-round regression, the 45min
  boundary, hour/day boundaries, the 30-day cliff, and the "412d ago"
  regression.
- **`web/messages/en.json`** — `relative.*` keys moved from `topics` namespace
  to `common` so all four pages share them.
- **`topics/page.tsx`** — removed local `formatRelative`, uses shared helper.
- **`prompts/page.tsx`** — removed local `formatRelative`, uses shared helper.
- **`insights/page.tsx`** — removed local `formatTimeAgo`, uses shared helper.
- **`traffic/page.tsx`** — removed local `timeAgo`, uses shared helper.

## Intentional behaviour change

Insights and Traffic previously showed timestamps older than 30 days as e.g.
`412d ago`. They now render as a locale date string (e.g. `15 Jul 2025`),
matching the existing behaviour of Topics and Prompts.

## Testing

- `pnpm tsc --noEmit` — 0 errors
- `pnpm lint` — 0 errors, 7 warnings all pre-existing
- `pnpm test` — 112/112 passing


gkhngyk:
The consolidation itself looks right — the floor-vs-round fix and the 30-day cliff are both pinned by tests, and moving the keys to `common` is the correct home for them.

Two small things before merge:

- `web/messages/en.json`: the closing curly quotes in `agent.emptyExample1` and `emptyExample2` turned into opening ones (`...doing?“`). Unrelated to this change and visible in the agent empty state — please restore the `”`.
- `topics/page.tsx` already declares `const common = useTranslations('common')` a line above the new `tCommon`; one of the two can go.

Maqbool61:
Both items addressed:

1. **`en.json`** — closing curly quotes (`"`) restored on `emptyExample1` and
   `emptyExample2`. The earlier replacement swapped both sides to opening quotes
   by mistake; confirmed via codepoint inspection (`0x201d` on both closing
   positions).

2. **`topics/page.tsx`** — removed the redundant `tCommon` declaration. The
   existing `common` variable was already in scope; call site updated to use it.

gkhngyk:
Both fixes verified and merged — the codepoint check on the quotes was the right instinct. Four copies down to one, with the boundaries pinned by tests. Thanks!

Issue #599, opened 2026-08-03
Citations: load failures leave the tables stuck on "Checking citation data…" forever


gkhngyk:
### Problem

When `getCitationsOverview` / `getCitationsTotal` reject, the Citations page's catch block is empty, so `hasAnyCitations` stays `null` — and the `null` branch of `EmptyRows` renders the "Checking citation data…" spinner. The result: on any load failure the tables spin forever with no error message and no retry, despite the catch comment claiming "user will see empty state / can retry".

### Evidence

- [`citations/page.tsx#L873-L877`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L873-L877) — `} catch { // swallow — user will see empty state / can retry`
- `hasAnyCitations` initialized `null` at [`#L777`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L777) and only set on success at [`#L856`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L856)
- Permanent spinner branch: [`#L447-L455`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/page.tsx#L447-L455)
- The URL detail page in the same feature already does this right with a `loadFailed` flag + error state + retry: [`citations/url/page.tsx#L159`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/%28dashboard%29/dashboard/citations/url/page.tsx#L159)

### Suggested fix

Mirror the URL detail page: set a `loadFailed` flag in the catch, render an error state with a retry button (re-invoking `loadData`), and make sure `EmptyRows` never receives an eternally-`null` `hasAnyCitations` after a settled fetch.

### Acceptance criteria

- A failed load shows an error message with a working retry, not an endless "Checking citation data…" spinner.
- Successful loads behave exactly as today.

PR #775, opened 2026-08-23
fix(citations): keep content visible during refetch

### LINKED ISSUE #656: Citations: keep content visible during refetch instead of swapping the whole page to a skeleton

BharadwajKanneveti:
## Summary

Updated the Citations page so existing content remains visible while citation data is being refetched.

Previously, every filter/date change replaced the page content with the full skeleton. Now:
- The skeleton is shown only during the initial load.
- During refetches, the existing data stays mounted.
- Existing content is dimmed and a loading spinner is displayed over it.
- Moved `brandHasCitations` out of `loadData` so it is fetched once per brand instead of on every filter change.

## Related issue

Closes #656

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx
@@ -856,6 +856,7 @@ export default function CitationsPage() {
   const [prompts, setPrompts] = useState<PromptOption[]>([]);
   const [availablePlatforms, setAvailablePlatforms] = useState<PlatformOption[]>([]);
   const [availableRegions, setAvailableRegions] = useState<string[]>([]);
+  const isRefetching = isLoading && data !== null;
 
   const activeBrandId = brand?.id ?? null;
 
@@ -916,13 +917,9 @@ export default function CitationsPage() {
     setIsLoading(true);
     setLoadFailed(false);
     try {
-      const [overview, hasAny] = await Promise.all([
-        getCitationsOverview(activeBrandId, gapFilters),
-        brandHasCitations(activeBrandId),
-      ]);
+      const [overview] = await Promise.all([getCitationsOverview(activeBrandId, gapFilters)]);
       if (stale()) return;
       setData(overview);
-      setHasAnyCitations(hasAny);
 
       // Surface filter options from the observed models/regions.
       const platformOptions = buildPlatformOptions(overview.rows);
@@ -952,6 +949,21 @@ export default function CitationsPage() {
       if (!stale()) setIsLoading(false);
     }
   }, [activeBrandId, gapFilters]);
+
+  useEffect(() => {
+    if (!activeBrandId) return;
+
+    let active = true;
+
+    brandHasCitations(activeBrandId).then((result) => {
+      if (!active) return;
+      setHasAnyCitations(result);
+    });
+
+    return () => {
+      active = false;
+    };
+  }, [activeBrandId]);
 
   useEffect(() => {
     loadData();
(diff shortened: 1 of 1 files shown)


gkhngyk:
The shape is right — the overlay matches the Insights pattern the issue pointed at, and hoisting `brandHasCitations` out of the refetch path is exactly the third acceptance criterion.

Two things to fix before merge:

**1. The KPI grid and the Sources card lose their spacing.** They used to be direct children of the page's `space-y-6` container (a fragment adds no DOM node); now they sit two divs deeper, so the 24px gap between them disappears. Insights solves it on the inner wrapper: `cn('space-y-6', isRefetching && 'opacity-60')`.

**2. A failing `brandHasCitations` brings back #599.** The new effect has no `.catch`, and the call now lives outside `loadData`'s try/catch — so a rejection is unhandled, `loadFailed` never flips, and `hasAnyCitations` stays `null`. `EmptyRows` reads that as "still checking" and both tables sit on the endless "Checking citation data…" spinner, which is the exact case the `loadFailed` branch was added for. Catching it and surfacing the failure (or at least settling the flag) closes that path.

Two smaller ones while you're in there: `const [overview] = await Promise.all([getCitationsOverview(...)])` can be a plain `await` now that it is a single call, and the `<>…</>` fragment inside the new wrapper is redundant — the class there can go through `cn` like the rest of the file.

gkhngyk (changes requested):
Marking this as requested changes so it doesn't look ready to merge — the two items above (the lost `space-y-6` on the new wrapper, and the uncaught `brandHasCitations` rejection leaving the tables on the endless "Checking citation data…" spinner) are the blockers. Ping me once they're in and I'll take another look.

BharadwajKanneveti:
@gkhngyk I have made the changes please check and let me know if you find any issues

gkhngyk (approved):
All four addressed, and the catch does the right thing — settling `loadFailed` is what keeps the tables off the endless "Checking citation data…" spinner. Thanks!

PR #782, opened 2026-08-23
feat: brand avatar fallback chain + logo URL field (fix #759)

### LINKED ISSUE #759: feat: let a brand set its own logo URL, and fall back to the site's favicon when Google has none
### Problem / motivation

A brand's icon comes from exactly one place. [`web/src/lib/favicon.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/favicon.ts) builds a URL pointing at Google's `s2/favicons` service, and that URL is stored as `brands.logo_url`. When Google has no icon for the domain, the brand has no icon — and there is nothing the user can do about it.

This is not rare — two brands added on the same day both landed on it. Both of those sites publish a perfectly good icon of their own; one serves a 180×180 `apple-touch-icon.png` with the actual logo, and Google simply hasn't indexed it. DuckDuckGo's equivalent service returns `404` for the same two domains, so swapping providers is not the answer either.

The second half of the problem is that the value is **derived on every save**. In [`brands/[id]/settings/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/brands/%5Bid%5D/settings/page.tsx):

```ts
const newLogoUrl = newPrimary ? getFaviconUrl(newPrimary.domain) : null;
```

So even correcting the row by hand is temporary — changing the primary domain recomputes it and the correction is gone.

### Proposed solution

Two parts. They are one issue because they touch the same value and the same call sites, but they can land in either order.

**1. Let a brand carry its own logo URL**

Add a logo URL field to brand settings. A plain URL text input — **no file upload and no storage bucket**, deliberately. `brands.logo_url` already exists and already holds a URL, so there is no migration.

When the field has a value, it is used and it **sticks**: saving domains must not overwrite it. The derived favicon is only a default for when the field is empty.

**2. Fall back to the site's own favicon when Google has nothing**

When the Google URL fails to load, try `https://<domain>/favicon.ico` before giving up and showing the initials.

Do this **at render time**, not by storing a second URL. The fallback is presentational; keeping it out of the database means `logo_url` continues to mean "the icon this brand chose" and no backfill is ever needed.

The natural shape is one small component — something like `BrandAvatar` — that walks the chain *manual URL → Google favicon → site `/favicon.ico` → initials* and is used everywhere a brand avatar appears today:

- [`components/layout/brand-switcher.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/layout/brand-switcher.tsx) (two places)
- [`components/dashboard/brand-card.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/dashboard/brand-card.tsx)
- [`components/providers/brand-guard.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/providers/brand-guard.tsx)
- [`brands/[id]/layout.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/brands/%5Bid%5D/layout.tsx)

**Notes**

- Radix's `AvatarImage` already reports a failed load, which is the hook the chain needs — you should not have to fetch anything yourself.
- Google answers a miss with **`404` carrying a valid PNG body** (the grey globe). Browsers will often render that rather than firing an error, so treating "image loaded" as success is not enough on its own. Check what actually happens for a known-missing domain before assuming which event fires.
- Don't validate the pasted URL by fetching it server-side. If it doesn't load, the chain already ends at the initials.
- `team-section.tsx` also uses `AvatarImage`, for member avatars. Leave it alone — different data, different fallback.

### Alternatives considered

**Upload the logo to storage.** The most robust option and explicitly not wanted: it drags in a bucket, size and format validation, and cleanup. A URL field solves the same problem with a text input.

**Fetch and cache the icon ourselves at brand creation.** Same objection, plus it makes brand creation depend on a third-party fetch succeeding.

**Switch favicon providers.** DuckDuckGo `404`s on the same two domains, so this trades one gap for another.

**Store the `/favicon.ico` fallback in `logo_url` instead of resolving it at render.** Then a site that later publishes a better icon stays stuck on the old one, and the column stops meaning one thing.

Maqbool61:
### Problem
Brand avatars had no fallback strategy — if `logoUrl` was missing or broken,
nothing rendered. Google's favicon service silently returns a grey-globe PNG
(not a 404) for unindexed domains, so `onError` alone couldn't catch it.
Additionally, changing the primary domain in the Domains tab would silently
overwrite any manually-set `logoUrl` in the database. Worse, `logo_url` was
being written at brand creation and on every primary-domain change with a
derived Google favicon URL, meaning 121 of 124 production brands carried a
derived URL — so the fallback chain never ran for them and the domain-change
guard was always false.

#### `BrandAvatar` component (`src/components/brand-avatar.tsx`)
New shared component that walks a 4-stage fallback chain:

1. **Manual `logoUrl`** — renders the URL set explicitly by the user
2. **Google favicon** — `https://www.google.com/s2/favicons?domain=...&sz=128`
3. **Site `/favicon.ico`** — direct fetch from the brand's primary domain
4. **Initials** — always-available `AvatarFallback` as the final safety net

Grey-globe detection: on `onLoadingStatusChange` (Base UI's correct hook —
not `onError` on the ``, which races with Base UI's internal preloader),
if `naturalWidth ≤ 16` after load the Google response is the 16×16 placeholder
globe — the component advances to stage 3 immediately without waiting for an
error that will never fire.

State resets cleanly via a `key`-based remount pattern when `logoUrl` or
`primaryDomain` changes — avoids calling `setState` inside `useEffect`.

All helpers (`resolveInitialStage`, `resolveUrl`, `advance`, `getInitials`,
`getPrimaryDomain`) are exported so tests import the real implementations and
break if the component changes.

#### Stop writing derived URLs at brand creation
Removed `getFaviconUrl(domain)` calls from both wizard creation paths:
- `onboarding/page.tsx:547`
- `brands/new/page.tsx:377`

`logo_url` now starts as `null` for new brands. The chain derives the favicon
at render time — nothing to store.

#### Drop the `logo_url` write in `DomainsTab` entirely
With the chain deriving the favicon at render time, there is nothing to guard.
The entire `logo_url` write block in `DomainsTab.handleSave` is removed.

#### One-off DB migration (`00073_clear_derived_logo_urls.sql`)
```sql
update brands set logo_url = null where logo_url like '%s2/favicons%';
```
Clears the 121 derived Google favicon URLs in production so existing brands
reach the fallback chain. The 2 manual URLs are unaffected (they don't match
the pattern). Safe to re-run — `WHERE` clause is idempotent.

#### Logo URL field in General Settings
Added a plain URL `<Input>` for `logoUrl` in the General tab. Uses `t('logoUrl')`
and `t('logoUrlHint')` — no hardcoded English. Validates that the value starts
with `https://` or `http://` before saving. Clearing the field stores `null`,
restoring favicon-derived behaviour.

#### `brand-logo-guard.ts` (`src/lib/brand-logo-guard.ts`)
Extracted `shouldUpdateLogoUrl` into a proper `lib/` module so tests import
the real predicate. The guard is no longer needed in `DomainsTab` (writes
dropped entirely) but the module and tests document the invariant.

#### Migrations — all four avatar sites migrated to `BrandAvatar`
- `src/components/layout/brand-switcher.tsx`
- `src/components/dashboard/brand-card.tsx`
- `src/components/providers/brand-guard.tsx`
- `src/app/[locale]/(dashboard)/dashboard/brands/[id]/layout.tsx`

`team-section.tsx` left alone as specified.

### Tests
65 unit tests across 3 files — no login or browser required:

| File | Tests | Covers |
|------|-------|--------|
| `src/components/brand-avatar.test.ts` | 35 | Fallback chain, stage transitions, grey-globe, `getFaviconUrl`, helpers — all imported from real source |
| `src/lib/brand-logo-guard.test.ts` | 13 | `shouldUpdateLogoUrl` — imported from real `lib/` module |

Closes #759

### DIFF
--- a/server/src/lib/anthropic-tracker.js
+++ b/server/src/lib/anthropic-tracker.js
@@ -4,6 +4,7 @@
  */
 
 import Anthropic from '@anthropic-ai/sdk';
+import { userLocationFor } from './locations.js';
 
 let client = null;
 
@@ -21,7 +22,9 @@ function getClient() {
  * Run a single prompt through Anthropic Messages API with web_search enabled.
  * @param {string} promptText
  * @param {string} model - e.g. "claude-sonnet-5"
- * @param {string} [region] - ISO 2-letter country code for web_search user_location
+ * @param {string} [region] - Location code for web_search user_location: an
+ *   ISO country (`DE`) or a US state (`US-CA`, #691). Country-only codes
+ *   produce exactly the payload they did before states existed.
  * @returns {Promise<{ text: string, citations: Array<{ url: string, title: string, startIndex: number, endIndex: number }>, model: string }>}
  */
 export async function runPromptAnthropic(promptText, model, region) {
@@ -33,8 +36,9 @@ export async function runPromptAnthropic(promptText, model, region) {
     max_uses: 5,
   };
 
-  if (region) {
-    webSearchTool.user_location = { type: 'approximate', country: region };
+  const userLocation = userLocationFor(region);
+  if (userLocation) {
+    webSearchTool.user_location = userLocation;
   }
 
   const response = await anthropic.messages.create({
--- a/server/src/lib/cloro-scraper.js
+++ b/server/src/lib/cloro-scraper.js
@@ -37,7 +37,8 @@ export function buildRequestBody(promptText, scraperId, region, state) {
   // State-level geo-targeting (#554): the AI endpoints accept a USPS state
   // code alongside country US and route through an in-state proxy. Google
   // AIO / AI Mode use location/uule for sub-country targeting instead and
-  // must never receive it — their branches below skip this spread.
+  // must never receive it — their branches below skip this spread, and since
+  // #691 the caller never even submits a per-state task for them.
   const stateField = country === 'US' && state ? { state } : {};
 
   if (scraperId === 'chatgpt-shopping') {
@@ -232,7 +233,8 @@ export function parseScraperResponse(result, scraperId) {
  * @param {string} scraperId
  * @param {string} [region]
  * @param {{ webhookUrl?: string, state?: string }} [opts] - `state` is the
- *   brand's optional USPS code (#554); only forwarded on US AI-endpoint tasks.
+ *   USPS code of the location this task targets (#554, per-prompt since
+ *   #691); only forwarded on US AI-endpoint tasks.
  * @returns {Promise<{ taskId: string, scraperId: string }>}
  */
 export async function submitScraperTask(promptText, scraperId, region, opts = {}) {
--- a/server/src/lib/locations.js
+++ b/server/src/lib/locations.js
@@ -1,0 +1,141 @@
+/**
+ * Tracking locations (#691).
+ *
+ * A prompt's `regions` column holds location codes, not plain countries: a
+ * bare ISO country (`DE`) or an ISO 3166-2 US state (`US-CA`). One code is
+ * one tracked location — one set of scraper and model calls, one plan-quota
+ * unit, one `prompt_results.region` value.
+ *
+ * Parsing lives here so exactly one definition decides what a code means.
+ * The web tier keeps a deliberately identical copy in `web/src/lib/region.ts`
+ * (same rule as the citation hostname split): both sides must agree on which
+ * codes are valid, or the app would offer targeting the worker cannot run.
+ */
+
+/**
+ * USPS code → full state name, for the AI providers' `user_location.region`,
+ * which takes a human-readable region rather than a code.
+ */
+export const US_STATE_NAMES = {
+  AL: 'Alabama',
+  AK: 'Alaska',
+  AZ: 'Arizona',
+  AR: 'Arkansas',
+  CA: 'California',
+  CO: 'Colorado',
+  CT: 'Connecticut',
+  DE: 'Delaware',
+  FL: 'Florida',
+  GA: 'Georgia',
+  HI: 'Hawaii',
+  ID: 'Idaho',
+  IL: 'Illinois',
+  IN: 'Indiana',
+  IA: 'Iowa',
+  KS: 'Kansas',
+  KY: 'Kentucky',
+  LA: 'Louisiana',
+  ME: 'Maine',
+  MD: 'Maryland',
+  MA: 'Massachusetts',
+  MI: 'Michigan',
+  MN: 'Minnesota',
+  MS: 'Mississippi',
+  MO: 'Missouri',
+  MT: 'Montana',
+  NE: 'Nebraska',
+  NV: 'Nevada',
+  NH: 'New Hampshire',
+  NJ: 'New Jersey',
+  NM: 'New Mexico',
+  NY: 'New York',
+  NC: 'North Carolina',
+  ND: 'North Dakota',
+  OH: 'Ohio',
+  OK: 'Oklahoma',
+  OR: 'Oregon',
+  PA: 'Pennsylvania',
+  RI: 'Rhode Island',
+  SC: 'South Carolina',
+  SD: 'South Dakota',
+  TN: 'Tennessee',
+  TX: 'Texas',
+  UT: 'Utah',
+  VT: 'Vermont',
+  VA: 'Virginia',
+  WA: 'Washington',
+  WV: 'West Virginia',
+  WI: 'Wisconsin',
+  WY: 'Wyoming',
+  DC: 'District of Columbia',
+};
+
+/**
+ * Scrapers with no sub-country targeting mechanism. Google AIO / AI Mode
+ * locate through location/uule parameters Cloro's AI endpoints don't accept,
+ * so a state code means nothing to them: they can only be run country-wide
+ * (#691, carried over from #554).
+ */
+const COUNTRY_ONLY_SCRAPERS = new Set(['google-aio', 'google-aimode']);
+
+/**
+ * Split a location code into what the providers actually take.
+ * `US-CA` → `{ country: 'US', state: 'CA' }`; `DE` → `{ country: 'DE',
+ * state: null }`. Unparsable or empty input yields a null country, which the
+ * callers treat as "no targeting" — the same as a prompt with no regions.
+ */
+export function parseLocation(code) {
+  if (typeof code !== 'string') return { country: null, state: null };
+  const trimmed = code.trim().toUpperCase();
+  if (!trimmed) return { country: null, state: null };
+
+  const [country, state] = trimmed.split('-');
+  if (!/^[A-Z]{2}$/.test(country || '')) return { country: null, state: null };
+  if (state === undefined) return { country, state: null };
+  // Sub-country targeting exists for US states only; anything else is
+  // treated as the bare country rather than silently sent as a bad code.
+  if (country === 'US' && US_STATE_NAMES[state]) return { country, state };
+  return { country, state: null };
+}
+
+/**
+ * The locations one scraper should actually be run for.
+ *
+ * Every code a state-capable engine can use survives as-is. For the
+ * country-only engines the list collapses to its distinct countries, so a
+ * prompt targeting California and Texas submits ONE country-wide Google task
+ * instead of two identical ones — which would have doubled the spend and
+ * written two rows differing only in a state label neither run honoured.
+ *
+ * Order is preserved so task submission stays deterministic.
+ */
+export function locationsForScraper(locations, scraperId) {
+  const list = Array.isArray(locations) && locations.length > 0 ? locations : [null];
+  if (!COUNTRY_ONLY_SCRAPERS.has(scraperId)) return list;
+
+  const seen = new Set();
+  const collapsed = [];
+  for (const code of list) {
+    const { country } = parseLocation(code);
+    const key = country ?? '';
+    if (seen.has(key)) continue;
+    seen.add(key);
+    collapsed.push(country);
+  }
+  return collapsed;
+}
+
+/**
+ * `user_location` for the API models' web-search tool, or null when the
+ * prompt is untargeted. `region` carries the full state name because that is
+ * what both providers document for the field — a USPS code is not a region
+ * name. Country-only locations produce exactly the payload they did before
+ * states existed, so untargeted and country-targeted runs are unchanged.
+ */
+export function userLocationFor(code) {
+  const { country, state } = parseLocation(code);
+  if (!country) return null;
+  const location = { type: 'approximate', country };
+  if (state) location.region = US_STATE_NAMES[state];
+  return location;
+}
--- a/server/src/lib/locations.test.js
+++ b/server/src/lib/locations.test.js
@@ -1,0 +1,74 @@
+import { describe, it, expect } from 'vitest';
+import { parseLocation, locationsForScraper, userLocationFor } from './locations.js';
+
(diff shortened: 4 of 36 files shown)


gkhngyk (changes requested):
Thanks for taking this one — the component shape is what the issue asked for, and all four avatar sites are migrated faithfully (`team-section.tsx` correctly left alone). Three things before it can land.

## 1. The manual/derived distinction doesn't exist in the data yet

The whole design rests on "`logo_url` empty → derive, `logo_url` set → the user chose it". Today `logo_url` holds a *derived* Google favicon URL for almost every brand: it's written at brand creation (`onboarding/page.tsx:547` and `brands/new/page.tsx:377`, both `getFaviconUrl(domain)`) and rewritten on every primary-domain change.

Measured on production: of 124 brands, **121 carry a derived Google URL, 2 a real manual URL, 1 is empty**. Two consequences:

- **The domain-change refresh stops.** `newPrimary?.domain !== oldPrimary?.domain && !brand.logoUrl` is false for those 121, so changing the primary domain leaves the old domain's icon in place. That's a regression of current behaviour.
- **The fallback chain never runs for them.** With `logoUrl` set, `initialStage` is `manual`, and the grey-globe check only fires in the `google` stage (`brand-avatar.tsx:104`). I asked the Google service for every distinct primary domain of those brands: **20 of 112 return the 16×16 grey globe today** — exactly the users this issue is for — and this PR leaves all 20 unchanged. The single brand with an empty `logo_url` is the only one that can reach the new path.

The fix is to make the column mean what the issue says it means ("the icon this brand chose") rather than to guess at runtime:

1. stop writing the derived URL at brand creation (the two wizard call sites);
2. drop the `logo_url` write in `DomainsTab` entirely — with the chain deriving the favicon at render time, there is nothing left to guard;
3. a one-off migration clearing the derived values (`update brands set logo_url = null where logo_url like '%s2/favicons%'` — every one of them is reproducible at render).

## 2. The new tests don't exercise the code they describe

Both suites re-implement the logic locally and test the copy:

- `brand-avatar.test.ts` defines its own `getInitials`, `resolveInitialStage`, `resolveUrl` and `advance` ("Replicate the pure helpers from brand-avatar.tsx"). The only thing imported from src is `getFaviconUrl`, which already existed.
- `brand-logo-guard.test.ts` defines `shouldUpdateLogoUrl` — there is no `web/src/lib/brand-logo-guard.ts` in the diff at all.

Delete `brand-avatar.tsx` and all 49 tests still pass. Please export the helpers from the component (or move them to a small `lib/` module) and import them, so the tests fail when the component changes.

## 3. `onError` is not the hook this codebase's avatar offers

The issue's note about Radix is out of date — `components/ui/avatar.tsx` wraps **Base UI** (`@base-ui/react/avatar`). Its `AvatarImage` preloads the src with its own `new Image()`, mounts the `` while the status is `loading`, and unmounts it as soon as that preload reports `error`. So whether your `onError` runs is a race between two handlers on the same request, and every icon is fetched twice. Base UI exposes the status directly:

```tsx
<AvatarImage onLoadingStatusChange={(status) => status === 'error' && advance()} … />
```

## The grey-globe heuristic itself checks out

Worth keeping as is. Measuring the service with `sz=128`: a miss answers **404 with a valid 16×16 PNG**, while real icons in our own data never come back that small — 48 at 128×128, 8 at 32×32, and the smallest seen is 28×28. Also, 14 of the responses were JPEG rather than PNG, so a format-based check would have been wrong; `naturalWidth` is the right signal. (The PR description says the miss is "not a 404" — it is one, browsers just render the body anyway.)

## Smaller things

- "Logo URL" and its help text are hardcoded English in a tab that otherwise goes through `t('brands.…')`. Please add the keys to `web/messages/en.json`.
- `type="url"` does nothing outside a form, so `example.com/logo.png` is stored as-is and resolves against the app's own origin. Trim and require a scheme.
- The checklist is in pnpm commands; the repo is yarn, and `.github`'s PR template should be filled in — see [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md).
- Optional, and a genuine gap: of those 20 grey-globe brands, only 3 publish a usable `/favicon.ico`. The site in the issue serves a 180×180 `apple-touch-icon.png` that Google hasn't indexed, which `/favicon.ico` alone won't find — one more stage before the initials would cover it.

Point 1 is the one that changes the shape of the PR; the rest are small next to it.

Maqbool61:
Thanks for the detailed review — all three points addressed:

**1. Manual/derived distinction doesn't exist in the data yet**
- Removed `getFaviconUrl(domain)` from both wizard creation paths
  (`onboarding/page.tsx:547`, `brands/new/page.tsx:377`) — `logo_url` now
  starts as `null` for new brands
- Dropped the entire `logo_url` write block from `DomainsTab.handleSave` —
  with the chain deriving the favicon at render time, there is nothing to store
  or guard
- Added migration `00073_clear_derived_logo_urls.sql`:
  `update brands set logo_url = null where logo_url like '%s2/favicons%'`
  — clears the 121 derived rows so existing brands reach the chain; the 2
  manual URLs are unaffected

**2. Tests don't exercise the code they describe**
- All helpers exported from `brand-avatar.tsx`
  (`resolveInitialStage`, `resolveUrl`, `advance`, `getInitials`, `getPrimaryDomain`)
- `shouldUpdateLogoUrl` extracted into `src/lib/brand-logo-guard.ts`
- Both test files import from the real sources — deleting the source breaks
  the tests

**3. `onError` is not the hook this codebase's avatar offers**
- Switched to `onLoadingStatusChange` — Base UI's documented API
- `onError` is completely gone from the component
- Grey-globe `naturalWidth ≤ 16` check preserved in the `loaded` +
  `google` stage branch

**Smaller things**
- `t('logoUrl')` and `t('logoUrlHint')` — no hardcoded English; both keys
  added to `web/messages/en.json`
- `type="url"` removed; inline error shown and save blocked when value
  doesn't start with `https://` or `http://`

gkhngyk (changes requested):
Thanks for the follow-up push — the migration is here and `brand-logo-guard.ts` now exists. Before the three points from the last round, there's something new that has to be fixed first.

## Blocker: this branch reverts #776 and #778

`167a516` restores the pre-#776 versions of three files. Against `main`:

| marker | main | this branch |
|---|---|---|
| `citation_url_candidates` / `citation_gap_domains` / `citation_competitor_sources` calls in `lib/actions/citations.ts` | present | **gone** |
| `scanFilteredResults` | removed | **back (3 uses)** |
| `CITATIONS_SCAN_PAGE_SIZE` | removed | **back (4 uses)** |
| the four `citation_*` RPC entries in `types/supabase.ts` | present | **deleted** |

Merged as-is, the Citations URL detail page and Competitor Gaps go back to paging every answer to the browser and to the 50,000-row cap that was silently hiding the newest answers from both. CI can't catch it — the reverted state is internally consistent, which is why every check is green.

Nothing in this PR needs those files, so the fix is to take `main`'s copies wholesale:

```bash
git fetch origin main
git checkout origin/main -- \
  web/src/lib/actions/citations.ts \
  'web/src/app/[locale]/(dashboard)/dashboard/citations/page.tsx' \
  web/src/types/supabase.ts
```

Afterwards `git diff origin/main --stat` should list only avatar/logo files plus the migration.

## 1. The manual/derived split — the migration alone doesn't do it

The migration is right, but it's one of the three pieces, and without the other two it's undone the moment anyone creates a brand:

- both wizards still write a derived URL at creation — `getFaviconUrl` is still called in `(onboarding)/dashboard/onboarding/page.tsx` and `dashboard/brands/new/page.tsx` (neither file is in the diff), so **every new brand gets `logo_url` set again** and never reaches the fallback chain;
- `DomainsTab` still writes `logo_url` on a primary-domain change (`settings/page.tsx:520-522`). With the chain deriving at render time there is nothing left for that write to do — please remove it rather than guard it.

Clear those two and the column finally means "the icon this brand chose", which is what the rest of the PR is built on.

## 2. The tests still test copies of the code

`brand-logo-guard.ts` exists now, but `brand-logo-guard.test.ts` doesn't import it — it still imports only `getFaviconUrl` and defines its own `shouldUpdateLogoUrl`. `brand-avatar.test.ts` still opens with "Replicate the pure helpers from brand-avatar.tsx", and `brand-avatar.tsx` still exports only `BrandAvatar`.

Deleting `brand-avatar.tsx` today leaves all the tests passing. Export the helpers (or move them into `lib/`) and import them in the tests.

## 3. Still `onError`

`brand-avatar.tsx:123` is `onError={advance}`. Base UI preloads the src itself and unmounts the `` when its own preload fails, so this stays a race and every icon is fetched twice:

```tsx
<AvatarImage onLoadingStatusChange={(status) => status === 'error' && advance()} … />
```

## Smaller

- `messages/en.json` is untouched — "Logo URL" and its help text are still hardcoded in a tab that otherwise uses `t('brands.…')`.
- `.trim()` is in, but a scheme still isn't required: `example.com/logo.png` is stored as-is and resolves against the app's own origin.
- `00073_clear_derived_logo_urls.sql` opens with `-- Migration: 00067_clear_derived_logo_urls`. Wrong number in the header.
- The description is still a pnpm checklist; the repo is yarn, and the `.github` PR template is still unfilled — see [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md).

The grey-globe heuristic and the component shape are still good — the work above is around them, not against them.

gkhngyk (approved):
Re-checked against `main` — everything from both rounds is addressed.

- **The #776/#778 revert is gone.** `lib/actions/citations.ts` and `types/supabase.ts` match `main`; `scanFilteredResults` and `CITATIONS_SCAN_PAGE_SIZE` are absent, and all four `citation_*` RPC calls and type entries are back. The diff is now avatar/logo files, the migration, `schema.sql` and `en.json` — nothing else.
- **Point 1.** Both wizards dropped `logoUrl` from their `createBrand`/`updateBrand` calls (the remaining `getFaviconUrl` uses are the on-screen previews, which is right), the domain-change write is gone from `DomainsTab` entirely, and the migration clears the derived values. The column now means what the design needs it to mean.
- **Point 2.** `brand-avatar.tsx` exports the helpers and the test imports them — deleting the component now breaks the suite, which was the point.
- **Point 3.** `onLoadingStatusChange` replaces `onError`.
- Smaller ones: `logoUrlHint` added and both keys resolve under the `brands` namespace, `handleSave` refuses a value without an `http(s)` scheme rather than only warning, and the migration header carries the right number.

Two cleanups before merge, both one-liners:

1. **`web/src/lib/brand-logo-guard.ts` is dead code.** Nothing under `web/src` imports `shouldUpdateLogoUrl` except its own test — dropping the `DomainsTab` write is exactly what made it unnecessary, and the file's own header says domain saves must leave `logo_url` untouched. Please delete it and `brand-logo-guard.test.ts`.
2. `"URL must start with https:// or http://"` is hardcoded next to `t('logoUrl')` and `t('logoUrlHint')` — it should be a message key too.

Deploy note for whoever merges: `00073` should be applied **after** the deploy. It nulls `logo_url` for 121 brands, and until the new component is live those brands render initials instead of their favicon — not breaking, but a visible window.

Nice work on the turnaround, and thanks for taking the citations revert seriously — that was the risky one.

PR #789, opened 2026-08-26
perf(visibility): check brand data existence with LIMIT 1 instead of an exact COUNT

### LINKED ISSUE #729: Visibility: replace the full-table count behind "does this brand have data" with an existence check
## Summary

Opening the Visibility page fires an exact `COUNT(*)` over the brand's `prompt_results` rows to answer a yes/no question: does this brand have any data at all? The answer is used as a boolean and nothing else, but the query scans every matching row to produce it.

On the largest brand in one production database that is a bitmap heap scan of 16,524 buffers — about 128 MB — to learn that the count is greater than zero. Warm it takes ~64 ms; cold it is 128 MB of disk reads, and it runs alongside seven other aggregate queries that all start in the same tick.

The count is the eighth of those eight parallel queries and the only one that reads the table directly rather than through an aggregate function. It is also the only one whose result is thrown away down to a single bit.

## Where it is

`getBrandResultsTotal` in `web/src/lib/actions/tracking.ts`:

```ts
const { count, error } = await supabase
  .from('prompt_results')
  .select('id', { count: 'exact', head: true })
  .eq('brand_id', brandId)
  .neq('platform', 'chatgpt-shopping');
```

Its only consumer, in `getInsightsData`:

```ts
const hasAnyData = unfilteredTotal !== null ? unfilteredTotal > 0 : summary.totalResults > 0;
```

The doc comment calls it a "cheap existence check ... counts, no rows". Counting no rows is not the same as reading no rows — `count: 'exact'` still visits every one of them.

It runs on the default page load, not only on an explicitly filtered view: the default date preset is `24h`, so `hasFilters` is true from the first render and `checkUnfiltered` is always set.

## Why it matters now

Statement timeouts on this workload arrive in clusters — seven to nine of them within the same second, one per query in the page's `Promise.all`. That is the whole page load failing together when the buffer cache is cold, which the user sees as a generic server-action error toast rather than as a slow page.

Removing this query does not fix the cold-cache problem on its own; the other seven still scan. But it is the largest single read of the eight, and the only one that can be deleted outright rather than optimized.

## Proposed change

Ask for one row instead of all of them:

```ts
const { data, error } = await supabase
  .from('prompt_results')
  .select('id')
  .eq('brand_id', brandId)
  .neq('platform', 'chatgpt-shopping')
  .limit(1);
```

and return a boolean. Measured on the same two brands:

| Brand | Rows | Current (exact count) | With `limit(1)` |
|---|---|---|---|
| Largest | ~53,000 | 16,524 buffers, 63.8 ms | 2 buffers, 3.8 ms |
| Smallest | 48 | — | 4 buffers, 1.7 ms |

The planner picks a plain index scan on `idx_prompt_results_brand_id` for a selective brand and a sequential scan with an early exit for a large one. Both stop at the first matching row.

One case is worth checking while implementing: a brand whose rows are all `chatgpt-shopping` has no early exit to reach, so the filter runs to the end of whatever plan is chosen. Confirm the shape there before merging.

Renaming the function to say what it now returns (`brandHasResults`) keeps the call site honest, and `unfilteredTotal !== null` becomes a straightforward boolean check.

## Scope

- `getBrandResultsTotal` and its single call site in `getInsightsData`
- The `hasAnyData` type stays a boolean, so no surface changes

## How to test

1. Open the Visibility page for a brand with a large result history and confirm the KPI cards, competitor table and Share of Voice render unchanged.
2. Apply a filter combination that returns nothing and confirm the page still distinguishes "no data for this filter" from "no data at all" — that distinction is the only thing this query exists for.
3. Open the page for a brand that has never completed a tracking run and confirm the empty state is the onboarding one, not the filtered one.

shahzainshafique:
## Summary

The Visibility page answered "does this brand have any data at all?" with an exact COUNT(*) over prompt_results (a ~128 MB bitmap heap scan on the largest brand) only to compare the total against zero. Since the one consumer reads it as a boolean, getBrandResultsTotal becomes brandHasResults, selecting id with .limit(1) so the planner stops at the first match (~16,500 buffers → 2). hasAnyData stays a boolean and the "no data for this filter" vs "no data at all" distinction is unchanged.

## Related issue

Closes #729 

## Type of change

- [ ] `feat` — New feature
- [ ] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/lib/actions/tracking.ts
+++ b/web/src/lib/actions/tracking.ts
@@ -574,18 +574,25 @@ async function getInsightsRecommendations(brandId: string): Promise<InsightsReco
   return { topics, prompts };
 }
 
-/** Cheap existence check for a brand's (non-shopping) prompt_results — counts, no rows. */
-async function getBrandResultsTotal(brandId: string): Promise<number> {
+/**
+ * Does this brand have any (non-shopping) prompt_results at all?
+ *
+ * A yes/no question, so it asks for a single row with LIMIT 1 and stops at the
+ * first match, rather than an exact COUNT that visits every matching row only
+ * to compare the total against zero.
+ */
+async function brandHasResults(brandId: string): Promise<boolean> {
   const supabase = await createClient();
-  const { count, error } = await supabase
+  const { data, error } = await supabase
     .from('prompt_results')
-    .select('id', { count: 'exact', head: true })
+    .select('id')
     .eq('brand_id', brandId)
-    .neq('platform', 'chatgpt-shopping');
-  // Throw rather than swallow: a failed count must not masquerade as "no data"
+    .neq('platform', 'chatgpt-shopping')
+    .limit(1);
+  // Throw rather than swallow: a failed check must not masquerade as "no data"
   // and silently flip the page to its empty state.
   if (error) throw new Error(error.message);
-  return count ?? 0;
+  return (data?.length ?? 0) > 0;
 }
 
 export interface InsightsData {
@@ -706,7 +713,7 @@ export async function getInsightsData(
     visibilityRate,
     filterOptions,
     recommendations,
-    unfilteredTotal,
+    unfilteredHasData,
   ] = await Promise.all([
     getInsightsSummary(brandId, filterOpts),
     getCompetitorComparison(brandId, filterOpts),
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Verified against production rather than taken on trust, and it holds: on the largest brand the count was 19,993 buffers / 81ms, the new check is **2 buffers / 0.085ms**.

The case I went looking for is the one that isn't in the issue. `LIMIT 1` has a planner trap on the *miss* — with no matching row and a seq scan, it reads to the end of the table. That is exactly the state this flag exists to detect: a brand with no data at all. So the change could have been slower precisely where it matters most.

It isn't. On a brand with zero rows the planner takes `idx_prompt_results_brand_created` and answers in 0.85ms. Checked a 48-row brand and a 21k-row brand too — index scans, all sub-millisecond. The seq scan only appears on the brand dense enough that the first heap page already contains a match.

Behaviour is unchanged: `getBrandResultsTotal` had exactly one consumer and it read the number as a boolean, the throw-don't-swallow guard is preserved, and the "no data for this filter" vs "no data at all" distinction still resolves the same way.

Two optional things, neither blocking — merge as is if you'd rather not touch it again:

- `unfilteredHasData !== null ? unfilteredHasData : summary.totalResults > 0` can now be `unfilteredHasData ?? (summary.totalResults > 0)`. The explicit null check was load-bearing while the value was a number, because `0` is falsy; narrowing it to a boolean retired that.
- The same shape lives in `traffic.ts`, where `countVisits()` is called with no arguments and read only as `> 0`. Out of scope here, and `ai_traffic_logs` is 6,108 rows today so there is nothing to gain yet — noting it so it isn't rediscovered from scratch later.

Nice, tightly scoped change. Thanks!

PR #673, opened 2026-08-07
fix: localize audit hub and report strings

BharadwajKanneveti:
## Summary

Localized audit hub and audit report UI strings by moving hardcoded English text into the `audit` translation namespace.

Changes include:
- Replaced hardcoded audit hub labels with `t(...)` translations.
- Added `audit.status.*` translations so audit statuses (`running`, `completed`, `failed`) render as user-friendly labels instead of raw enum values.
- Localized audit report labels including signals, recommendations, findings, fixes, copy actions, and category sections.
- Removed user-visible hardcoded strings from audit-related components.

## Related issue

Closes #670 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix
- [ ] `chore` — Maintenance / dependencies
- [ ] `docs` — Documentation only
- [ ] `refactor` — Code change that neither fixes a bug nor adds a feature
- [ ] `test` — Adding or updating tests

## Checklist

- [x] Branch follows the naming convention (`feature/`, `fix/`, `chore/`, `docs/`) — see https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] `yarn lint` passes (run from `web/`)
- [x] `yarn typecheck` passes (run from `web/`)
- [x] `yarn format:check` passes (run from `web/`)
- [x] Changes are focused — one concern per PR

PR #790, opened 2026-08-26
fix(audit): confirm deletion from recent audits

### LINKED ISSUE #614: Site Audit: deleting from the recent-audits list has no confirmation (detail page has one)

BharadwajKanneveti:
## Summary

Add a confirmation dialog before deleting an audit from the recent-audits list.

Previously, clicking the trash icon immediately triggered the delete action. Now, clicking the trash icon opens a confirmation dialog, and the audit is deleted only after the user explicitly clicks **Delete**. Cancelling leaves the audit unchanged.

The existing translated status labels for non-completed audits are already present, so no additional changes were needed for that requirement.

## Related issue

Closes #614 

## Type of change

- [ ] `feat` — New feature
- [x] `fix` — Bug fix

## Checklist

### DIFF
--- a/web/src/app/[locale]/(dashboard)/dashboard/audit/page.tsx
+++ b/web/src/app/[locale]/(dashboard)/dashboard/audit/page.tsx
@@ -13,6 +13,16 @@ import {
   AlertCircle,
   RefreshCw,
 } from 'lucide-react';
+import {
+  Dialog,
+  DialogClose,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
+} from '@/components/ui/dialog';
 import { Link, useRouter } from '@/i18n/navigation';
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Button } from '@/components/ui/button';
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Faithful mirror of the detail page's dialog — same title, description, outline Cancel and destructive confirm, and no new copy since `deleteAudit` / `deleteConfirm` / `cancel` were already there. That was the whole ask.

The thing I went looking for was the one-dialog-per-row choice, since the issue asked not to render N dialogs if the list can be long. It holds here on both counts: the list is capped at 50 server-side (`server/src/routes/audits.js:361`), and `DialogContent` renders through `DialogPortal`, so a closed dialog mounts its trigger and nothing else. Fifty triggers is not a cost worth `pendingId` state to avoid.

You also stayed clear of the trap in this particular row: the trigger is a **sibling** of the `<Link>`, not inside it. Nested, a click would have opened the dialog and navigated to the detail page at the same time.

And you read the second half of the issue correctly — the raw `running` / `failed` status label was fixed in #673 and is already going through `t('status.${h.status}')` on main, so there was nothing left to do there.

Nothing blocking. One observation not worth changing: the dialog title and the confirm button are both "Delete audit", which reads a little repetitive — but the detail page does the same, and matching it is the right call over being locally prettier.

Thanks for picking this one up.

BharadwajKanneveti:
Hi @gkhngyk , this is unrelated to the PR, but could I DM you about something when you have a moment? I’m not sure which platform you’re most active on. Can you please tell me?

PR #795, opened 2026-09-02
feat: replace browser confirms with destructive dialogs

### LINKED ISSUE #756: feat: confirm destructive actions with the Dialog pattern instead of the browser's confirm()
### Problem / motivation

Destructive actions are confirmed three different ways depending on where you happen to be.

**A `Dialog`** — [`audit/[id]/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/audit/%5Bid%5D/page.tsx) wraps the delete button in a `Dialog` with a title, a description, Cancel, and a `destructive` confirm button.

**A typed confirmation** — deleting a brand asks you to type the brand name (`settings.deleteConfirmPlaceholder` in `web/messages/en.json`).

**The browser's own `confirm()`** — four places:

| File | Action |
|---|---|
| [`agent/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/agent/page.tsx) `deleteConversation` | delete a conversation |
| [`reports/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/reports/page.tsx) `handleDelete` | delete a report |
| [`settings/api-keys-section.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/api-keys-section.tsx) `handleRevoke` | revoke an API key |
| [`settings/agent-section.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/agent-section.tsx) `handleClear` | remove the org's Anthropic key |

`confirm()` is the browser's own modal. It ignores the theme, cannot be styled, cannot mark the destructive choice as destructive, renders as an OS alert on mobile, and blocks the whole tab while it is open. Two of the four also have their prompt text hardcoded in English while the other two use `t()`.

The last two are the ones that sting: revoking an API key breaks every client using it, and clearing the org key cuts the whole team off from the agent. Those deserve better than a grey system box.

### Proposed solution

Move all four to the `Dialog` pattern the audit page already uses. That page is the reference — copy its shape rather than inventing a new one:

- `DialogTitle` naming the action, `DialogDescription` saying what it costs
- `DialogClose` with `variant="outline"` to cancel
- `DialogClose` with `variant="destructive"` to confirm

**Notes**

- `agent` and `reports` delete a specific row from a list, so the dialog needs to know *which* one. A single dialog driven by a `pendingId` piece of state reads better than one dialog rendered per row — pick whichever you can make clean, but don't render N dialogs if the list can be long.
- The existing `t('deleteConfirm')` strings already say the right thing; reuse them as the `DialogDescription`.
- The two settings sections have their text inline. While you're there, move those strings into `web/messages/en.json` alongside the others.
- Keep the disabled/pending state the handlers already manage (`deletingId`, `clearing`) so the confirm button can't be double-clicked.

**Out of scope**

Brand deletion's typed confirmation stays as it is. Typing the name is a deliberately higher bar for a deliberately bigger action.

### Alternatives considered

**Add an `AlertDialog` primitive.** The usual shadcn split gives destructive confirmations their own component. There isn't one in `web/src/components/ui/` today, and adding a primitive is a larger change than this needs — `Dialog` already does the job on the audit page.

**Leave `confirm()` alone.** It does technically confirm. But the app looks like an application everywhere else and like a 1998 browser here, and the two key-related actions are exactly where a user should be able to read carefully before clicking.

**Related:** #614 covers a delete that has *no* confirmation at all. Different problem, no overlap.

BharadwajKanneveti:
## Summary

Replaced the browser's native `confirm()` dialogs for destructive actions with the existing `Dialog` pattern used across the dashboard.

This updates confirmation flows for:
- Deleting conversations
- Deleting reports
- Revoking API keys
- Removing the organization's Anthropic API key

The new dialogs use the application's styling and translations, with explicit Cancel and destructive confirmation actions. Shared dialogs are used for list-based actions so we don't render one dialog per row.

## Related issue

Closes #756 

## Type of change

- [x] `feat` — New feature
- [ ] `fix` — Bug fix

## Checklist


gkhngyk (changes requested):
Good approach — one shared dialog per list instead of one per row is the right call, and all four `confirm()` call sites are gone with no path left unconfirmed (`deleteConversation`, `handleDelete` and `handleClear` each have exactly one caller now).

One thing blocks merge.

## `cancel` doesn't exist in these namespaces

All four dialogs call `t('cancel')`, but `t` is bound to `agent`, `reports` and `settings` — and `cancel` only lives in `common`:

```
namespaces with cancel:     common, audit, content, competitors
namespaces without:         agent, reports, settings, + 12 more
```

`src/i18n/request.ts` sets no `getMessageFallback`, so next-intl's default applies and each button renders the key path instead of the word: **`agent.cancel`**, **`reports.cancel`**, **`settings.cancel`**, with a `MISSING_MESSAGE` error in the console.

Three of the four files already bind a common translator a line or two away:

| File | Already in the file | This PR |
|---|---|---|
| `settings/api-keys-section.tsx` | `tCommon('cancel')` (L268) | `t('cancel')` (L224) |
| `dashboard/reports/page.tsx` | `tc('cancel')` (L473) | `t('cancel')` (L286) |
| `settings/agent-section.tsx` | `tCommon` bound at L53 | `t('cancel')` (L220) |
| `dashboard/agent/page.tsx` | — | `t('cancel')` (L377), needs a `useTranslations('common')` |

`yarn typecheck` can't catch this: the project doesn't declare next-intl's `IntlMessages` augmentation, so message keys aren't checked at build time.

## Also worth fixing

**The Anthropic key warning lost its point.** It read *"Remove the Anthropic API key? Team members will lose access to the agent until a new key is saved."*; `settings.agent_removeKeyConfirm` now says *"The agent will no longer be able to use it."* The key is organization-wide, so the team-impact sentence is the part that made it a warning — please keep it.

**`reports.delete` duplicates `common.delete`.** That file already binds `tc`, so `tc('delete')` covers it without a new key. (`settings.remove` and `settings.revoke` are fine — `common` has neither.)

## Optional

- The destructive button is wrapped in `DialogClose`, so the dialog dismisses before the request resolves. That makes the `disabled={clearing}` and `disabled={deletingId === pendingConfirm}` props you added unreachable, and a failed delete surfaces only as a toast. `dashboard/competitors/page.tsx` (L1403–1428) keeps its confirm dialog open with a spinner instead — worth matching.
- In the agent dialog the title and the confirm button are the same string ("Delete conversation"). Reports gets this right: "Delete Report" as the title, "Delete" on the button.
- `agent/page.tsx`: the Dialog import block landed below `const AgentChart = dynamic(...)`, splitting the imports.
- `Delete Report` / `Revoke API Key` — most titles in `en.json` are sentence case ("Reset your password", "No competitors yet", "Remove {name}?").

gkhngyk (commented):
Everything from the last round is in — the four `cancel` lookups, the restored key-removal warning, the dropped `reports.delete`, the titles, the import placement. Thanks.

The fix introduced one new thing.

### `handleClear` no longer resets `confirmClear`

In `settings/agent-section.tsx` the confirm button stopped being a `DialogClose`, so nothing sets `confirmClear` back to `false` any more — and `handleClear` doesn't either.

It's hidden right now because the dialog sits inside `{configured && (…)}`: a successful clear flips `configured` to false and unmounts the whole block, so the dialog disappears. But the state stays `true`.

1. Remove the key → confirm → success. `confirmClear` is still `true`.
2. Paste a new key on the same page and hit Save.
3. `handleSave` → `load()` → `configured` is true again → the block remounts → `<Dialog open={confirmClear}>` → the "Remove key" confirmation opens by itself, right after the key was saved.

No reload involved, it's all client state. `setConfirmClear(false)` in `handleClear`'s `finally` covers it — `reports/page.tsx` already does the equivalent with `setPendingConfirm(null)`.

### Non-blocking

The destructive button in `dashboard/agent/page.tsx` and `settings/api-keys-section.tsx` is still wrapped in `DialogClose`, so two of the four dialogs close instantly and two now wait with a spinner. Both behave correctly; worth settling on one if you want them uniform.

Fix the first one and this is good to merge.

gkhngyk (approved):
`45cf84b` closes it. `setConfirmClear(false)` in the `finally` is the right spot — it covers the failure path too, which matches how the dialog behaved before this PR.

Re-checked the whole thing at head:

- No native `confirm()` or `alert()` left anywhere in `web/src` — all four call sites converted.
- All four Cancel buttons resolve through `common`, and every new message key is referenced exactly once. No dead entries, and `reports.delete` is gone in favour of `common.delete`.
- The two dialogs that still close via `DialogClose` reset their state through `onOpenChange`, so none of the four can be left holding a stale `open`.

CI is green including the production build. Approving — thanks for staying with it through three rounds.

PR #804, opened 2026-09-06
fix(prompts): keep examples within schema bounds

### LINKED ISSUE #797: bug: the prompt generator's own example is shorter than the schema it must satisfy
## Description

`POST /api/prompts/from-topics` validates what the model returns with this schema — [`server/src/routes/prompts.js#L594`](https://github.com/ansvisor/ansvisor/blob/main/server/src/routes/prompts.js#L594):

```js
const topicPromptSchema = z.object({
  topicPrompts: z.array(
    z.object({
      topic: z.string(),
      prompts: z.array(z.string().min(30).max(100)).length(5),
    }),
  ),
});
```

The system prompt in the same handler then offers the model this example:

> Write them like real quick searches, e.g. "best CRM for small business" or "top email marketing tools <year>".

`best CRM for small business` is **27 characters** — three short of the `min(30)` that the same request enforces. The model is being shown a sample it is not allowed to produce. The second example is exactly 30 characters, so it sits right on the boundary.

`generateObject` rejects the entire object if any single prompt falls under 30 characters, and the call is wrapped in `withRetry(..., { attempts: 3 })`. A model that follows the example closely therefore burns all three attempts and the endpoint returns 500. Onboarding lands the user on the prompts step with nothing generated and shows `Couldn't generate prompt suggestions right now`.

## Steps to reproduce

Count the characters in the first example string in the system prompt and compare it to `min(30)` in `topicPromptSchema`, a few lines above. Failures are intermittent by nature — they depend on how literally the model copies the example — which is exactly why this is worth removing as a source of noise.

## Expected behavior

Every example the system prompt shows the model satisfies the schema that will validate the model's answer.

## Suggested fix

Either lengthen both examples so they sit comfortably inside 30–100 characters, or lower `min` to match them — either is fine, as long as the two agree.

Then add a unit test that asserts each example string embedded in the system prompt passes the schema's own bounds, so the prompt and the schema cannot drift apart again. Tests live beside the code (`server/src/lib/*.test.js`) and run with `yarn test` from `server/`.

copacabanaservice01:
## Problem

The `/api/prompts/from-topics` schema requires every generated prompt to contain 30–100 characters, but its first example was only 27 characters. A model following that example could therefore produce an otherwise valid response that Zod rejects and the route retries.

## Approach

- Replace both examples with natural 46-character queries that sit comfortably inside the existing bounds.
- Add a focused regression test that reads the examples and the schema bounds from the route source, resolves the dynamic year, and checks every example against those bounds.

## Tests

- `vitest run src/routes/prompts.test.js`
- `vitest run` — 29 files, 405 tests passed
- `eslint .`
- Prettier verification for both changed files
- `git diff --check upstream/main...HEAD`

## Risk

Low. This changes only instructional example text; the schema, API response, retry behavior, dependencies, authentication, and database behavior are unchanged.

## Exclusions

- No schema-bound changes
- No retry or provider changes
- No unrelated prompt cleanup

Closes #797

### DIFF
--- a/server/src/routes/prompts.js
+++ b/server/src/routes/prompts.js
@@ -630,7 +630,7 @@ router.post('/from-topics', async (req, res) => {
 
 RULES:
 - Keep prompts SHORT and concise — between 30 and 100 characters. Aim for 40-80 characters.
-- Write them like real quick searches, e.g. "best CRM for small business" or "top email marketing tools ${new Date().getFullYear()}".
+- Write them like real quick searches, e.g. "best CRM software for growing small businesses" or "top email marketing tools for startups in ${new Date().getFullYear()}".
 - Do NOT use the brand name "${brandName}" in the prompts.
 - Each prompt should be relevant to the specific topic it belongs to.
 - Focus on queries where the brand could organically appear in AI-generated answers.
--- a/server/src/routes/prompts.test.js
+++ b/server/src/routes/prompts.test.js
@@ -9,8 +9,11 @@ describe('topic prompt examples', () => {
     const schema = source.match(
       /const topicPromptSchema[\s\S]*?prompts: z\.array\(z\.string\(\)\.min\((\d+)\)\.max\((\d+)\)/,
     );
+    expect(schema).not.toBeNull();
     const bounds = { min: Number(schema[1]), max: Number(schema[2]) };
-    const examples = source.match(/e\.g\. "([^"]+)" or "([^"]+)"/).slice(1);
+    const rule = source.match(/- Write them like real quick searches, e\.g\. ([^\n]+)$/m);
+    expect(rule).not.toBeNull();
+    const examples = [...rule[1].matchAll(/"([^"]+)"/g)].map(([, example]) => example);
     const currentYear = String(new Date().getFullYear());
     const resolvedExamples = examples.map((example) =>
       example.replace('${new Date().getFullYear()}', currentYear),
--- a/server/src/routes/prompts.test.js
+++ b/server/src/routes/prompts.test.js
@@ -1,0 +1,24 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+const source = readFileSync(fileURLToPath(new URL('./prompts.js', import.meta.url)), 'utf8');
+
+describe('topic prompt examples', () => {
+  it('keeps every documented example within the prompt schema bounds', () => {
+    const schema = source.match(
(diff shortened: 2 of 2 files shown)


copacabanaservice01:
Hi maintainers — this fix is ready for review. The available checks are green and the branch is mergeable. When you have a chance, could you please review it and merge if it looks good? I’m happy to address any feedback.

gkhngyk (dismissed):
Verified at `bec7493` in a clean worktree: full server suite green (29 files, 405 tests), eslint and prettier clean — matches what you listed. Both examples land at 46 characters, inside the 30–100 bounds and inside the "aim for 40-80" band the same rule asks for.

I also tried to break the test, and the cases that matter hold: putting the old 27-character example back fails it with `expected 27 to be greater than or equal to 30`, and tightening the schema to `min(50)` fails it with `expected 46 to be greater than or equal to 50`. Reading the bounds out of the source instead of hardcoding them is what makes that second one work — good call, that's the part that keeps the two from drifting apart again.

One gap worth closing while you're here.

### A third example slips through

`/e\.g\. "([^"]+)" or "([^"]+)"/` captures exactly two quoted strings. Add a third example to that sentence and it goes unchecked — I appended `or "bad"` (3 characters) and the test still passed. Since the test's job is "every documented example satisfies the schema", matching every quoted string on the rule line closes it:

```js
const rule = source.match(/- Write them like real quick searches[^\n]*/)[0];
const examples = [...rule.matchAll(/"([^"]+)"/g)].map((m) => m[1]);
```

That also anchors the match to this specific rule, so a future `e.g. "…" or "…"` elsewhere in the file can't be picked up by mistake.

While you're in there: both `match()` calls throw `Cannot read properties of null` if the source drifts. Wrapping the schema across lines, for instance, gives `TypeError: … (reading '1')` rather than anything naming what moved. An `expect(...).not.toBeNull()` before each read makes that eventual failure explain itself.

Neither of these touches the fix, which is right as it stands.

copacabanaservice01:
Thanks for the careful break-test. I pushed 8c2efe4: the test now anchors to the quick-search rule line, collects every quoted example with `matchAll`, and checks both source matches before indexing. I reproduced the gap first—the old test passed with a third 3-character example; the updated test rejects it. The clean focused test, full server suite (405 tests), lint, and Prettier check all pass.

gkhngyk (approved):
Verified locally: the third-example gap is closed (`expected 3 to be greater than or equal to 30`), and both null guards now fail readably instead of throwing. 405 tests, lint and format green.

Thanks for the quick turnaround.

PR #806, opened 2026-09-11
test(web): add unit tests for getActiveEngineIdsForPlan

### LINKED ISSUE #800: test: cover getActiveEngineIdsForPlan with unit tests
## Description

`getActiveEngineIdsForPlan` in [`web/src/lib/plan-engines.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/plan-engines.ts) has no test coverage. It is pure — plan id in, engine and model id lists out — and it decides which scrapers and models a customer's prompts are tracked against, so getting it wrong either under-tracks a paying customer or spends scraper budget a plan doesn't cover.

Its central rule is a three-way distinction that is easy to collapse into two by accident:

- `allowedScrapers` / `allowedModels` **absent** on the plan → everything is allowed (Growth, Enterprise, Self-hosted).
- An **array**, even an **empty** one → only the listed ids are allowed. Starter ships `allowedModels: []`, which must resolve to no models at all, not to every model.
- An unknown or missing plan id falls back to `starter`.

## Suggested fix

Add `web/src/lib/plan-engines.test.ts` covering each plan defined in [`web/src/config/plans.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/config/plans.ts), the empty-array case, and the `null` / `undefined` / unknown-id fallback.

Scope this to `getActiveEngineIdsForPlan` only. The other export in the file, `alignPromptsToPlanForOrg`, talks to the database and needs mocking — worth a separate issue rather than folding it in here.

The repo already uses Vitest with a dozen sibling examples — [`web/src/lib/region.test.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/region.test.ts) is a good short template. Run with `yarn test` from `web/`.

TianHengZhuang:
## What

Adds `web/src/lib/plan-engines.test.ts` covering `getActiveEngineIdsForPlan`, whose central three-way rule (`allowedScrapers` / `allowedModels` absent → everything allowed; an array, even empty → only the listed ids; unknown/missing plan id → `starter`) had no test coverage. Closes #800.

## Coverage

| Case | Expectation |
| --- | --- |
| `starter` (cloud) | `allowedModels: []` resolves to **no** models, only the two listed scrapers |
| `growth` (cloud) | every scraper, still no API models |
| `enterprise` (cloud) | every scraper, no plan-level models (`plan_overrides` are layered on downstream by `alignPromptsToPlanForOrg`) |
| `self_hosted` (cloud) | every scraper and every model |
| `undefined` / `null` / `''` / `'not-a-plan'` | same as `starter` (fallback) |
| self-hosted instance, non-self-hosted id | id is ignored, everything allowed |
| `NEXT_PUBLIC_IS_CLOUD` unset | defaults to self-hosted |
| return value | fresh arrays, so callers cannot mutate the plan config |

## Notes

- Scoped to the pure helper only, as suggested: `alignPromptsToPlanForOrg` talks to the database and deserves its own issue.
- Importing the module also pulls in the service-role Supabase client, which throws at import time when `SUPABASE_SERVICE_ROLE_KEY` is unset (as in CI). The test stubs `@/lib/supabase/admin` with `vi.mock` so only the gating logic is under test; `vi.stubEnv` drives the cloud/self-hosted branch.
- Expected ids are derived from `ALL_SCRAPERS` / `ALL_MODELS` in `@/config/prompt-options` rather than hard-coded, so the suite keeps passing when the catalog grows.

## Verification

```
npx vitest run src/lib/plan-engines.test.ts   # 8 passed
npx prettier --check src/lib/plan-engines.test.ts  # All matched files use Prettier code style!
yarn test / yarn format:check                 # green
# Note: I installed deps with `npm install`, which resolves newer supabase
# typings than web/yarn.lock and surfaces 4 unrelated errors in
# src/lib/actions/* under tsc --noEmit. CI's `yarn install --frozen-lockfile`
# path is unaffected by that drift, and none of the 4 files are touched here.
```

### DIFF
--- a/web/src/lib/plan-engines.test.ts
+++ b/web/src/lib/plan-engines.test.ts
@@ -1,0 +1,116 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { ALL_MODELS, ALL_SCRAPERS } from '@/config/prompt-options';
+import { getActiveEngineIdsForPlan } from './plan-engines';
+
+// This module also exports the Supabase-backed alignPromptsToPlanForOrg, so
+// importing it pulls in the service-role client — which throws at import time
+// when NEXT_PUBLIC_SUPABASE_URL / SUPABASE_SERVICE_ROLE_KEY are unset, as they
+// are in CI. Only the pure gating helper is under test, so a stub is enough.
+vi.mock('@/lib/supabase/admin', () => ({ supabaseAdmin: {} }));
+
+const ALL_SCRAPER_IDS = ALL_SCRAPERS.map((s) => s.id);
+const ALL_MODEL_IDS = ALL_MODELS.map((m) => m.id);
+
+const STARTER_SCRAPER_IDS = ['chatgpt-web', 'perplexity-web'];
+
+/**
+ * getPlan() short-circuits to the self-hosted plan unless the instance is
+ * cloud, so every cloud expectation has to set the flag the helper reads.
+ */
+function useCloud(isCloud: boolean) {
+  vi.stubEnv('NEXT_PUBLIC_IS_CLOUD', isCloud ? 'true' : 'false');
+}
+
+describe('getActiveEngineIdsForPlan', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  /**
+   * The rule under test (#800): an absent `allowedScrapers` / `allowedModels`
+   * means every engine is allowed, while an array — even an empty one — means
+   * only the listed ids are.
+   */
+  it('gives Starter only the scrapers it lists and no API models', () => {
+    useCloud(true);
+
+    expect(getActiveEngineIdsForPlan('starter')).toEqual({
+      platforms: STARTER_SCRAPER_IDS,
+      models: [],
+    });
+  });
+
+  it('gives Growth every scraper but still no API models', () => {
+    useCloud(true);
+
+    expect(getActiveEngineIdsForPlan('growth')).toEqual({
+      platforms: ALL_SCRAPER_IDS,
+      models: [],
+    });
+  });
+
+  it('leaves Enterprise API models to plan_overrides', () => {
+    useCloud(true);
+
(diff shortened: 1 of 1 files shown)


gkhngyk (approved):
Verified locally at `c65f050`: 207 tests pass, and typecheck, lint and prettier are clean with the lockfile's dependencies — the four `tsc` errors you hit come from the `npm install` drift you described and are gone under `yarn install --frozen-lockfile`, so nothing to chase there.

I also tried to break the source to see whether the tests actually hold:

| Mutation | Result |
|---|---|
| Treat an empty `allowedModels` as "everything allowed" | 3 tests fail |
| Invert the scraper filter | 2 tests fail |
| Return a shared array instead of a fresh one | 3 tests fail |

So the distinction the issue was about — an empty array meaning *none*, not *all* — is genuinely pinned now.

Covering `getPlan`'s self-hosted short-circuit goes past what the issue asked for and is the right call: that branch decides what a self-hosted instance tracks and had nothing holding it. Deriving the expectations from `ALL_SCRAPERS` / `ALL_MODELS` keeps the suite honest as the catalog grows, and the `vi.mock` of `@/lib/supabase/admin` is necessary rather than incidental — that module throws at import time without a service-role key.

### Optional

`defaults to self-hosted when NEXT_PUBLIC_IS_CLOUD is unset` depends on the ambient environment not carrying that variable. I checked, and Vitest does not load `web/.env` into `process.env`, so it passes on a normal checkout — I copied a real `.env` in to confirm. It would still fail for anyone who exports `NEXT_PUBLIC_IS_CLOUD=true` in their shell. `vi.stubEnv('NEXT_PUBLIC_IS_CLOUD', undefined)` deletes the variable in Vitest 3 and makes the test independent of the environment it runs in.

Unrelated to this PR: the `planId ?? 'starter'` inside `getActiveEngineIdsForPlan` is dead code — `getPlan` already falls back twice. I removed it and all eight tests still passed. Nothing to change here; noting it in case it is worth a follow-up.

Issue #796, opened 2026-09-03
bug: a failed topic save is reported as a prompt-generation failure


gkhngyk:
## Description

In the onboarding wizard, `handleGeneratePrompts` saves the selected topics and generates prompts for them inside a single `try` block — [`web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx#L657`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(onboarding)/dashboard/onboarding/page.tsx#L657):

```ts
try {
  // Save topics to DB
  await createTopics(createdBrand.id, topicNames);

  const res = await fetch(`${AEO_SERVER_URL}/api/prompts/from-topics`, { ... });
  ...
} catch (err) {
  setPromptGenError(true);
  setTopicPrompts(topicNames.map((topic) => ({ topic, prompts: [] })));
  setStep(4);
}
```

`createTopics` ([`web/src/lib/actions/topic.ts#L17`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/topic.ts#L17)) deletes the brand's existing topics before inserting the new ones. If that insert fails, two things happen at once:

1. The brand is left with **zero** topics — the delete has already run.
2. The catch treats the failure as a prompt-generation failure: the user sees `Couldn't generate prompt suggestions right now` and is advanced to step 4.

From step 4 the user can type prompts by hand and finish onboarding. The prompts save successfully, because a prompt row carries its topic as a plain `category` string, so nothing appears to be wrong — but the `topics` table is empty for that brand, and every topic-level report is silently blank from then on.

## Steps to reproduce

1. Start onboarding and reach the topics step.
2. Make the topic insert fail — e.g. temporarily `throw new Error('boom')` at the end of `createTopics`.
3. Select topics and click Continue.

## Expected behavior

The wizard stays on the topics step and reports that **saving the topics** failed, so the user can retry rather than continuing on top of a brand that has none.

## Actual behavior

The wizard advances to step 4 and blames prompt generation.

## Suggested fix

Give the topic save its own `try`/`catch` and its own error state, ahead of the prompt-generation call — a `topicSaveError` rendered alongside the existing `promptGenError` message is enough, no restructuring of the step needed.

Keep the current behavior for a genuine *prompt-generation* failure: advancing to step 4 with empty per-topic groups is deliberate (#379), so that one transient model failure doesn't strand the user with nothing to do.

[...]

PR #807, opened 2026-09-14
refactor: extract shared AI suggestion fetches into ai-brand-setup.ts

### LINKED ISSUE #803: refactor: the three AI suggestion fetches are written twice
## Problem / motivation

The onboarding wizard and the "add a brand" page ask the API server for the same three things — suggested topics, prompts generated from those topics, and suggested competitors — and each has its own near-identical copy of the fetch code:

| Call | Onboarding wizard | Add-a-brand page |
|---|---|---|
| `POST /api/topics/suggest` | `fetchTopicSuggestions` (~L589) | `fetchTopicSuggestions` (~L446) |
| `POST /api/prompts/from-topics` | `handleGeneratePrompts` (~L657) | `handleGeneratePrompts` (~L516) |
| `POST /api/competitors/suggest` | `fetchCompetitorSuggestions` (~L772) | `fetchCompetitorSuggestions` (~L611) |

Files: [`(onboarding)/dashboard/onboarding/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(onboarding)/dashboard/onboarding/page.tsx) (1,748 lines) and [`(dashboard)/dashboard/brands/new/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/brands/new/page.tsx) (1,334 lines).

Each copy repeats the same four steps: read the Supabase session, build the `Authorization` header, POST a body assembled from the brand fields, and check `res.ok`. That is six places to keep in step, and changes to how the browser reaches the API have had to be applied to every one of them — #793 was exactly that.

## Proposed solution

Extract the three calls into one module — a `useAiSuggestions` hook, or three plain async functions in `web/src/lib/` — that owns session lookup, headers, and error shape, and have both pages call it.

Keep this a pure refactor: same requests, same request bodies, same user-visible behaviour on both pages. The two pages differ in what they do with the results (the wizard advances a step, the brand page does not), so the extracted layer should return data and throw on failure, leaving the per-page handling where it is.

`web/src/app/[locale]/(dashboard)/dashboard/brands/[id]/prompts/page.tsx` also calls `/api/prompts/from-topics`, but with a different body and a different flow around it. Leaving it out is fine; fold it in only if it fits the extracted signature without contorting it.

## Alternatives considered

Leaving the duplication in place. It has been survivable so far, but every fix to this path currently has to be made twice, and the onboarding funnel is where a missed copy costs the most.

## Notes

Run `yarn lint`, `yarn typecheck`, and `yarn build` from `web/`, then walk both flows end to end — onboarding from step 1, and adding a second brand from the dashboard. See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md).

Maqbool61:
Closes #803

## What this does

Extracts the three browser-side AI suggestion POSTs that were duplicated
across the onboarding wizard and the add-a-brand page into a single shared
module: `web/src/lib/ai-brand-setup.ts`.

**Before:** six copies of the same four steps (session lookup → Authorization
header → POST → res.ok check) spread across two 1000+ line files.

**After:** three plain async functions in one place. Both pages import and
call them; all per-page side effects (state updates, step advances, analytics
tracking) stay exactly where they were.

## Files changed

| File | Change |
|---|---|
| `web/src/lib/ai-brand-setup.ts` | **New.** Owns `suggestTopics`, `generatePromptsFromTopics`, `suggestCompetitors` |
| `web/src/app/[locale]/(onboarding)/dashboard/onboarding/page.tsx` | Replaced 3 inline fetch blocks with calls to the shared helpers |
| `web/src/app/[locale]/(dashboard)/dashboard/brands/new/page.tsx` | Same |

Net: −114 lines.

## Pure refactor

Same requests, same request bodies, same user-visible behaviour on both
pages. No changes to error handling shape, no new dependencies.

## Testing

- `yarn lint` — 0 errors ✅
- `yarn typecheck` — 0 errors ✅
- `yarn build` — fails with `supabaseUrl is required` on `main` before and
  after this change (pre-existing env-var issue on local builds without
  `.env.local`, not introduced here) ✅


gkhngyk (approved):
Verified this behaves identically. The part I cared about most is that the proxy survived, so I checked the built client bundle rather than only the source: `BROWSER_API_BASE_URL` resolves to `/api/aeo` in the same chunks that hold the three calls, and that is the base they are built from. The absolute API origin does still appear in the bundle, but only inside `getPublicApiBaseUrl`, which is deliberately absolute for the tracking snippet.

All six call sites compared one by one: request bodies, headers, the `res.ok` check, the thrown messages, and every per-page side effect (state updates, `setStep`, `track`) are unchanged. Keeping `createTopics` inside the same `try` is the right call here — that one belongs to #796, not to a refactor — and leaving the third `from-topics` caller out of scope matches what the issue suggested.

Locally: 199 tests, typecheck, lint, prettier and `next build` all clean. The `supabaseUrl is required` you hit is only the missing env vars; the placeholder build CI runs passes.

### Optional

`BrandSetupParams` makes `website` required, but two of the three endpoints never send it — callers pass it to all three and it is silently dropped in two. The behaviour is correct; the type just promises something the requests do not keep. Making it optional, or splitting the shape per call, would say what actually happens.

`getAuthHeader()` produces `"Bearer undefined"` when there is no session, and the server answers 401. The inline copies did the same, so this is not a regression — but a shared helper is the natural place to tighten it, and `lib/actions/volumes.ts` already sets the precedent with `throw new Error('Not authenticated')`.

Issue #802, opened 2026-09-03
chore: move the Content area's toast strings into messages/en.json


gkhngyk:
## Problem / motivation

The app is wired for translation with `next-intl`, and [`web/messages/en.json`](https://github.com/ansvisor/ansvisor/blob/main/web/messages/en.json) already holds ~800 keys across 19 namespaces. But 82 `toast.error(...)` / `toast.success(...)` calls across the dashboard still pass hardcoded English literals, so the moment a second locale is added they stay untranslated with nothing to signal it.

This issue covers the **Content area only**, as a self-contained first slice — 14 strings across three files, all of which already have a `t` bound to the `content` namespace, so no new wiring is needed:

| File | Strings |
|---|---|
| `web/src/app/[locale]/(dashboard)/dashboard/content/page.tsx` | 6 |
| `web/src/app/[locale]/(dashboard)/dashboard/content/[id]/page.tsx` | 5 |
| `web/src/app/[locale]/(dashboard)/dashboard/content/_webhook-settings.tsx` | 3 |

Examples: `Failed to load content opportunities`, `Sent to workflow!`, `Webhook URL is required`.

## Proposed solution

Move each literal into `web/messages/en.json` under the `content` namespace (`content.webhook` for the webhook settings file, which is the namespace that file already uses) and replace the call site with `t('...')`.

Keep the wording identical — this is a mechanical move, not a copy rewrite. Where the same sentence appears in more than one file, use one shared key rather than duplicating it.

## Alternatives considered

Doing all 82 at once. Splitting by feature area keeps each PR reviewable and lets several people work in parallel; if this slice goes smoothly the same treatment can be repeated for the remaining areas.

## Notes

Run `yarn format:check`, `yarn lint`, and `yarn build` from `web/` before opening the PR — `messages/*.json` is covered by Prettier. See [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md).

PR #817, opened 2026-09-15
Closed #796

bernalalexis-try:
Separated topic saving from prompt generation in handleGeneratePrompts, each with its own try/catch. If saving topics fails, the user stays on that step with a retry message, instead of advancing as if prompt generation had failed.


gkhngyk (approved):
This does exactly what the issue asked. The topic save gets its own `try`, a failure keeps the user on the topics step with a message they can act on, and the prompt-generation branch keeps the #379 behaviour of advancing with empty groups — which was the part worth preserving, since one transient model failure should not strand anyone.

Checked the details that matter here:

- The early return resets `loadingPrompts` itself, so the button recovers even though it leaves before the `finally` below.
- `setTopicSaveError(false)` at the top means a retry clears the banner rather than stacking on it.
- `setStep(4)` stays out of the topic-save path entirely — the brand can no longer end up with zero topics while the wizard advances.
- The banner matches the three sibling error banners on this page, down to the amber border and the phrasing.

Verified locally: 216 tests, typecheck, lint and prettier all clean.

### Optional

The message is a hardcoded string while the rest of the page is too, so this is consistent rather than a regression — but if #802's move of toast strings into `messages/en.json` reaches this page, this one wants to go with it.

Thanks for picking this up.

gkhngyk (approved):
Exactly the shape we framed — nice first contribution.

Verified against main: the two failure paths are properly separated. A failed topic save now keeps the user on the topics step with its own message (flag reset at handler start, loading cleared before the early return, so retry works cleanly), while the topics-saved-but-generation-failed path keeps today's advance-to-step-4 behavior from #379 untouched. The error box matches the existing amber pattern and sits where the eye already looks. CI is green.

Two small notes for next time, neither blocking:

- Branch naming: we use a `fix/` prefix rather than working on the fork's `main` — see [CONTRIBUTING](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md).
- The PR template (Summary / Related issue / How to test) helps reviews go faster; we'll fix up the squash title on merge this time.

Thanks!

PR #822, opened 2026-09-17
fix(a11y): add aria-label to remaining icon-only buttons

### LINKED ISSUE #798: bug: icon-only buttons have no accessible name
## Description

Seven buttons in the dashboard render nothing but an icon and carry no `aria-label`, `title`, or visually hidden text. A screen reader announces them with no name, so there is no way to tell what they do.

| File | Line |
|---|---|
| `web/src/app/[locale]/(dashboard)/dashboard/prompts/_fanout-tab.tsx` | 684 |
| `web/src/app/[locale]/(dashboard)/dashboard/prompts/_suggestions-card.tsx` | 547 |
| `web/src/app/[locale]/(dashboard)/dashboard/prompts/_suggestions-card.tsx` | 562 |
| `web/src/app/[locale]/(dashboard)/dashboard/topics/_suggestions-card.tsx` | 274 |
| `web/src/app/[locale]/(dashboard)/dashboard/topics/_suggestions-card.tsx` | 289 |
| `web/src/app/[locale]/(dashboard)/dashboard/brands/[id]/topics/page.tsx` | 193 |
| `web/src/components/layout/mobile-nav.tsx` | 48 |

This is the minority case: 24 of the app's 31 `size="icon"` buttons already carry an accessible name, so there are plenty of nearby examples to copy.

## Steps to reproduce

Open any of the files above and look for an accessible name on the listed button, or navigate the page with a screen reader and tab to it.

## Expected behavior

Every icon-only control announces what it does.

## Suggested fix

Add an `aria-label` to each of the seven buttons, describing the action rather than the icon ("Remove topic", not "Trash icon").

Where the file already calls `useTranslations`, add the string to [`web/messages/en.json`](https://github.com/ansvisor/ansvisor/blob/main/web/messages/en.json) under the namespace that file uses and reference it with `t(...)`, matching the surrounding style instead of hardcoding a literal.

Run `yarn lint` and `yarn build` from `web/` before opening the PR — see [CONTRIBUTING.md](https://github.com/ansvisor/ansvisor/blob/main/CONTRIBUTING.md).

Maqbool61:
Closes #798

## What

Adds accessible names to the remaining icon-only buttons that had none.
A full-codebase audit was also run to catch any unlisted cases.

### `web/src/components/layout/mobile-nav.tsx`
Added `aria-label={t('toggleMenu')}` directly on the `<Button>` inside
the `render` prop of `<SheetTrigger>`. The existing `<span
className="sr-only">` was also updated to use the same translation key
instead of a hardcoded string.

### `web/src/components/settings/team-section.tsx`
Two icon-only buttons were missing accessible names (not listed in the
issue but caught during the audit):
- **`MemberRow`** — trash button that opens the "Remove team member"
  dialog: added `aria-label={t('removeMember')}`
- **`InvitationRow`** — overflow button that opens the invitation actions
  dropdown: added `aria-label={t('invitationActions')}`

Both sub-components also received their own `useTranslations('settings')`
call, since they are separate named functions and do not share the hook
from the parent.

### `web/messages/en.json`
Added three new keys:
- `nav.toggleMenu` → `"Toggle navigation menu"`
- `settings.removeMember` → `"Remove team member"`
- `settings.invitationActions` → `"Invitation actions"`

## What was already fixed

Six of the seven buttons listed in the issue already carried an
`aria-label` in `main` at the time this PR was opened:

| File | Line | Status |
|---|---|---|
| `prompts/_fanout-tab.tsx` | 684 | ✅ already fixed |
| `prompts/_suggestions-card.tsx` | 547 | ✅ already fixed |
| `prompts/_suggestions-card.tsx` | 562 | ✅ already fixed |
| `topics/_suggestions-card.tsx` | 274 | ✅ already fixed |
| `topics/_suggestions-card.tsx` | 289 | ✅ already fixed |
| `brands/[id]/topics/page.tsx` | 193 | ✅ already fixed |
| `mobile-nav.tsx` | 48 | 🔧 fixed in this PR |

## Testing

- `yarn lint` — 0 errors
- `yarn build` — compiled and type-checked successfully
- `yarn format:check` — all files pass Prettier
- Manual: navigate the affected pages with a screen reader or browser
  accessibility inspector and confirm each button announces its action


gkhngyk (approved):
Verified against main: the six "already fixed" rows check out, and the audit claim holds — the two `icon-sm` buttons here were the only remaining unnamed icon-only controls (dialog/sheet close buttons already carry sr-only text). Namespace choices and the per-component `useTranslations` calls are right. Nice thorough work — thanks!

PR #825, opened 2026-09-18
feat: replace inert Project tab with working Organization tab

### LINKED ISSUE #753: feat: replace the inert Project settings tab with a working Organization tab
### Problem / motivation

Settings has a **Project** tab that does nothing.

[`web/src/app/[locale]/(dashboard)/dashboard/settings/page.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/app/%5Blocale%5D/(dashboard)/dashboard/settings/page.tsx) renders two inputs and a Save button under it. The inputs have no `value` and no `onChange`, the button has no `onClick`, and no action is imported for them. Nothing is read, nothing is written — clicking Save silently does nothing, which reads to the user as "saved".

"Project" also isn't a thing in this product. The model is **organization → brand**; there is no project table. So the tab is both inert and named after a concept that doesn't exist.

Meanwhile the organization's own name has no home in the UI at all. It is written once during onboarding by `createOrganization` in [`web/src/lib/actions/organization.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/organization.ts) and after that there is no way to see it or change it. A typo at signup is permanent. The only place `organizations.name` surfaces is the invitation page ("Join {organizationName}"), so an invitee sees the org name but a member never does.

The same page has a second dead Save button: the **Account** tab binds `displayName` to state but never persists it.

### Proposed solution

Replace the Project tab with an **Organization** tab that works.

**Show**
- Organization name — editable
- Slug — read-only (it is referenced elsewhere; renaming it is out of scope here)
- Plan and member count — read-only, both already available via `getTeamInfo` in [`web/src/lib/actions/team.ts`](https://github.com/ansvisor/ansvisor/blob/main/web/src/lib/actions/team.ts)

**Rename**
- Add an `updateOrganizationName` server action next to `createOrganization`.
- **The action must check the caller's role itself and reject anyone who is not `admin`** — do not rely on the database policy alone.
- Trim the input, reject empty, and cap the length (the column is `text`; something like 100 chars is plenty).
- Leave `slug` untouched — changing it is a separate concern with its own consequences.
- Non-admins should see the name, not an editable field.

**While you're in there**
- Wire the Account tab's Save button to persist `full_name` to the user's profile, or remove the button. A button that does nothing is worse than no button.

**Notes**
- `TeamSection` in [`web/src/components/settings/team-section.tsx`](https://github.com/ansvisor/ansvisor/blob/main/web/src/components/settings/team-section.tsx) is the closest existing pattern: load on mount, `toast` on success and failure, skeleton while loading. Follow it rather than inventing a new shape.
- The tab id `project` appears in the `Section` union and in the nav list; both need renaming. Check for `?tab=` deep links before changing the id.
- Strings live in `web/messages/en.json` under `settings` — `project` and `projectName` are there and will need replacing.
- No migration is needed. `organizations.name` already exists.

### Alternatives considered

**Delete the tab and do nothing else.** That removes the false affordance but leaves the org name unreachable, which is the more useful half of the fix.

**Put the rename inside the Team tab.** Team is about members and invitations; org identity is a different concern and the sidebar already has room for its own entry.

**Make the slug editable too.** Tempting, but the slug is used as a stable identifier and changing it needs its own thought about what breaks. Deliberately out of scope.

Maqbool61:
Closes #753

## What

Replaces the dead "Project" settings tab (inputs with no handlers, a Save
button that did nothing) with a fully working "Organization" tab, and wires
the Account tab's Save button which had the same dead-button bug.

### `web/src/lib/actions/organization.ts`

Three new server actions:

**`updateOrganizationName(name)`**
- Calls `requireAdmin()` (belt-and-braces on top of RLS, mirrors `team.ts`)
- Trims whitespace, rejects empty, caps at 100 chars
- Updates `organizations.name` only — leaves `slug` untouched

**`getOrganization()`**
- Returns `{ name, slug, plan, memberCount }` for the current org

**`updateAccountName(name)`**
- Writes `profiles.full_name` (the column all app queries read)
- Syncs `auth.users.user_metadata.full_name` so `supabase.auth.getUser()`
  reflects the new name immediately without waiting for a session refresh

### `web/src/components/settings/organization-section.tsx` (new)

Follows `TeamSection` / `AgentSection` patterns exactly:
- `useCallback` load in `useEffect`, `Skeleton` while loading
- `canAdmin` gate via `useUserRole()` — non-admins see all fields read-only
  with a lock hint; admins get an editable name field
- Save button disabled when name is unchanged or empty
- `toast.success` / `toast.error`, `Loader2` spinner on the button
- Slug always read-only with an explanatory hint

### `web/src/app/.../settings/page.tsx`

- `'project'` removed from the `Section` type; replaced with `'organization'`
- `OrganizationSection` imported and rendered for the new tab
- `updateAccountName` wired to the Account Save button with loading state
  and `toast` on success/error
- Legacy deep-link `?tab=project` preserved — still maps to `organization`
- Removed the now-dead `Label` / `Input` / `Button` block for the old
  Project card (Label import kept — still used in the Account tab)

### `web/messages/en.json`

- Removed unused keys: `project`, `projectName`, `domain`
- Added: `organization`, `org_title`, `org_description`, `org_name`,
  `org_namePlaceholder`, `org_slug`, `org_slugHint`, `org_plan`,
  `org_members`, `org_adminOnly`, `org_loadFailed`, `org_saveFailed`,
  `accountSaveFailed`

## Verification

- `yarn lint` — 0 errors
- `yarn build` — compiled and type-checked successfully (`supabaseUrl` error
  is a missing local env var, not related to these changes)
- `yarn format:check` — all files pass Prettier


gkhngyk (changes requested):
The Organization tab and Account Save work is exactly what #753 asked for — requireAdmin, validation, slug untouched, auth-metadata sync all check out, and the component follows the TeamSection pattern well.

One blocking issue: the en.json cleanup removed more than the three Project-tab keys. `signalTexts.visibility_slipping`, `states.slipping`, `actionTexts.protect_visibility`, and the `diagnose_slip` / `review_responses` / `reinforce_content` task keys are all live — they're referenced through dynamically composed keys (`` t(`${kind}.title`) ``) in `web/src/lib/signals/display.ts` and `web/src/lib/action-center/display.ts`, and the server emits those kinds in production (`server/src/lib/pulse/metrics.js`, `server/src/lib/action-center/generate.js`). An unused-key scan can't see dynamic composition, which is probably how they got flagged. Please restore those blocks — keeping only the `project` / `projectName` / `domain` removals — and this is good to go.

Maqbool61:
Fixed. The en.json was rebuilt from the latest main as the base, so visibility_slipping, protect_visibility, diagnose_slip, review_responses, and reinforce_content are all restored. Only project, projectName, and domain were removed. Force-pushed.

gkhngyk (approved):
Verified the update: the en.json cleanup is now scoped to the three Project-tab keys only, with all the dynamically referenced actionCenter keys intact, and the code is unchanged from the version already reviewed. Thanks for the quick turnaround — good to go.

Issue #818, opened 2026-09-16
Action Engine — build plan and phase tracker


gkhngyk:
Tracking issue for building out the Action Engine, from the V1 Action Engine specification and its companion Task Library specification.

It exists so the state of this work is answerable without reading the code or asking anyone: each phase has an **exit condition** that is measured, not judged, and the marker below moves as phases land.

## 📍 Where we are

**Phases 1, 2, 2.5 and 3a are done. Phase 3b is next.**

| | Now | Target |
|---|---|---|
| Action Definitions | **6** | 62 |
| Signal detectors registered | **13** | enough to trigger each enabled definition |
| Detectors that have actually fired | **10** | all of them |
| Protect family definitions | **1** | 10 |
| Outcome measurement | **built, waiting on its first window** | every closed action measured |
| Task planning | fixed list per kind | context-aware plan |
| Candidate layer | none | aggregating queue |

### Verified in production

The nightly run after #823 landed confirmed three phases at once:

- **`sharp_drop` fired for the first time in its existence** — a 72% relative visibility loss that the old 15-point threshold would have passed over in silence. One signal across the whole customer base, which is the volume the calibration predicted.
- **Phase 1's central fix held.** A signal detected two weeks after its action was opened attached to that action as evidence, logged a `signals_linked` event, and left the action's target scope untouched. No second action for the same kind; no rewrite of work someone may already be executing against.
- **Phase 2 correctly declined to measure.** The one closed action shut two days ago, and its after-window has not elapsed. Waiting is the behaviour — measuring a window still in progress was the bug caught during that phase.

Already shipped: the Action Center itself (#808), its write-policy fix (#813), task editing (#814), the History tab (#815).

## The phases

Phases run in order. One PR per phase, and a phase is not started until the previous one's exit condition holds — then we look at the data before moving on.

### Phase 1 — Data model correctness  ✅ #819

The generator gives every action `dedup_key = kind`, so a brand holds at most one action per kind, forever. New signals overwrite the open action's payload, and a completed action is resurrected by setting its status back to `new`. That breaks four invariants at once: target scope expands silently, history is mutated, a definition cannot produce a second cycle, and `no_improvement` stores an outcome in the status column.

- [x] Partial unique index so the one-active-per-definition rule applies to open actions only
- [x] Remove the reopen branch — a closed action is immutable, a new cycle is a new row
- [x] Freeze an open action's target snapshot; new signals no longer rewrite it
[...]

PR #827, opened 2026-09-20
feat(action-center): make a definition a file, not a line in the engine

gkhngyk:
## Summary

Phase 4 of #818. Definitions leave the engine.

The eight action definitions lived in one array inside `generate.js`, with
their payload shaping in a `switch` beside it. That is fine at eight and
impossible at the sixty-two the plan calls for: every new definition means
editing the matching engine, and a mistake in one breaks the others.

Each definition is now a file in `server/src/lib/action-center/definitions/`,
carrying its own category, signal kinds, task list, payload shaping, version,
and the data a brand must have for it to apply. The registry discovers them by
listing the directory and validates each on load. The engine no longer knows
any definition by name.

Adding a definition is now: add a file.

### Discovery over a manifest

A manifest of imports is one more place to forget, and forgetting it fails
silently — the definition simply never fires. A directory listing plus strict
validation fails loudly instead, at startup, with the filename in the message:
an unknown category, a version that is not a positive integer, a source that
does not exist, a missing payload function. A test asserts the rest — unique
ids, reproducible order, no two definitions consuming the same signal kind,
and that every signal kind a definition listens for is one a detector actually
emits. That last one is the quiet failure worth catching: a definition naming
a kind nothing produces would simply never fire, and nothing would say so.

### Eligibility

A definition declares the sources it requires and the ones that merely enrich
it. `sources.js` answers, per brand, what is actually available: tracking
(always), competitors, site audits, an analytics connection.

Measured over the live database before writing it:

| source | brands | of |
|---|---|---|
| tracking | 151 | 151 |
| competitors | 126 | 151 |
| site audits | 20 | 151 |
| analytics | 7 | 151 |

And against the actions already raised: **none of the 28 would have been
blocked by eligibility.** That is the expected result and the reason the check
is worth having anyway. A brand with no competitors never receives a
competitor signal, so the definition would not have matched regardless — but
today that is a coincidence of how the detectors happen to work, and the
declaration makes it a contract. Phase 5's task planner reads the same answer
to give two brands different plans for the same action kind, which is why it
is computed once here rather than inferred twice.

Eligibility gates opening a cycle, never maintaining one: a brand that
disconnects an integration mid-cycle keeps the action it was already given.

### The daily cap

Five new actions per brand per day. It binds on nothing today — eight
definitions produce at most three on a brand's busiest day, because a brand
rarely has more than a couple of distinct conditions firing at once:

| new actions in one day | brand-days |
|---|---|
| 1 | 12 |
| 2 | 5 |
| 3 | 2 |

The ceiling is the definition count, and the plan takes that to sixty. At that
point a bad night hands someone a list nobody reads, and the engine talks its
way out of being believed. Anything over the line is not lost: the condition
is still firing, so the slot reopens tomorrow, in priority order — impact
first, then weight of evidence. Phase 6 replaces that ordering with scoring
across candidates; with eight definitions there is nothing yet for a score to
say that this does not.

The budget is the day's, not the run's, and it is counted from rows the
generator already fetches rather than from a separate query. An action raised
this morning and dismissed by lunch still spent its slot — the brand was told
about it.

### Definition versions

`actions.definition_version` (migration `00086`) records which version of a
definition raised an action. Nothing reads it yet; recording it costs a
column, and reconstructing it later would cost the history. Without it,
History re-reads its own past against today's rules every time a definition
is tuned — the same reasoning as resolving a task's text at creation (`00083`).

### Web: an unknown kind is shown, not hidden

`getActions` and `getActionHistory` filtered out any row whose `kind` this
build did not recognise. That filter was the last thing making a server-side
definition useless until the web shipped too, and it was hiding a worse
problem: an unrecognised kind would have crashed the icon lookup and the text
switch rather than skipping them.

An action this build has no copy for is now presented from its id and its
family — `expand_platform_visibility` reads as "Expand platform visibility",
the icon falls back to the category's, and the description and goal come from
one generic string. Written copy still wins wherever it exists. Retiring the
filter is what lets the two sides move on separate release cycles, which is
the whole point of the registry.

## Related issue

Refs #818

## Type of change

- [x] `feat` — New feature

## How to test

```
cd server && yarn test          # 502 tests, 37 new
cd web && yarn test && yarn typecheck && yarn lint && yarn build
```

The behaviour to look for on the next nightly run: the `[actions] generated`
log line now reports `ineligible` and `capped` alongside `created`,
`refreshed` and `resting`, and new rows carry `definition_version`.

Migration `00086` is applied to the hosted database.

**Based on `feature/gap-signals` (#826), which is not merged yet** — the
registry covers all eight definitions including the two that PR adds. GitHub
will retarget this to `main` once #826 lands.

## Checklist

- [x] Branch follows the naming convention (`feature/`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SQ2H6mSd9JDpmqonk2VWBj


gkhngyk:
Superseded by #828. This branch was based on `feature/gap-signals` and merged into it rather than into `main` — by then #826 had already been squash-merged, so the commit landed on a branch that no longer leads anywhere and none of this reached `main`. #828 is the same commit cherry-picked onto `main`.

